PR #70, opened 2015-04-18
Use ServiceLoader to discover backends

EronWright:
Here's a PR for the improvement to use ServiceLoader.

Please note:

1. Jocl backend is not yet supported.   
2. `org.nd4j.linalg.jcublas.SimpleJCublas::init` was changed to throw an exception if initialization fails.

To add support for a new backend (see also [ServiceLoader javadoc](http://docs.oracle.com/javase/7/docs/api/java/util/ServiceLoader.html)):

1. Extend the service type `org.nd4j.linalg.factory.Nd4jBackend` with a backend-specific implementation (e.g. `org.nd4j.linalg.jblas.JblasBackend`).
2. In the implementation, return an appropriate priority value.   Highest priority wins.  A pair of constants (BACKEND_PRIORITY_CPU, BACKEND_PRIORITY_GPU) establish a convention.
3. Implement `isAvailable` to self-assess whether the backend is usable.
4. Add a provider-configuration file in the backend's resource directory `META-INF/services`, referring to the service implementation.


nd4j-jcublas-parent/nd4j-jcublas-common/src/main/java/org/nd4j/linalg/jcublas/SimpleJCublas.java
```diff
@@ -97,7 +97,7 @@ public static void init() {
         try {
             KernelFunctionLoader.getInstance().load();
         } catch (Exception e) {
-            e.printStackTrace();
+            throw new RuntimeException(e);
```
EronWright: Please note the semantic change here.

agibsonccc:
This is a good start. I'd be curious to see how this can fit in with opencl as well. Basically, opencl has no blas bindings itself. I'm going to need submodules for each kind of backend (mkl,cublas,amd,..)

My thinking on this would be to also use this for discovering which kind of provider is used for blas as well. This plays well with what I would have liked to see. I had forgotten about META-iNF/services.

It's been a while since I looked at osgi and some of the other kinds of service discovery. (I've also played with JNDI as well).

I'll be updating the jocl backend over the next day or 2. Only a few more details left there. After that I can look at integrating this.

agibsonccc:
Ok after looking at opencl. I'm going ot use netlib to link against the different blas impls in the different opencl backends. Merging this.

PR #96, opened 2015-05-09
improve multithreaded gpu support

agibsonccc:
Support Jcublas2. Have had problems with linking against cublas2 files. Would like help with testing.


agibsonccc:
Problems linking against jcublas2 when invoked. UnsatisfiedLinkError cublas v1 works perfectly.

PR #110, opened 2015-05-15
Updated dependencies for jocl 1.9 to common and removed Mojo.

nyghtowl:


nyghtowl:
@agibsonccc - check this out

agibsonccc:
Decided I'm going to create a dedicated gpu module for sharing logic between jocl and jogl. This should give us some flexibility with executing kernels and the like.

PR #147, opened 2015-06-05
avoid unnecessary boxing

EronWright:


EronWright:
Not tested this, just eyeballed it.  During a recent debugging session, I caught a thread in the midst of creating a boxed Double but scratched my head about whether it was intentional.  Assumedly not.

PR #164, opened 2015-06-17
fix serialization bug (long->int)

EronWright:
Fixes regression from 1efb7ae5e78600401e3c0fed79ffc7425a2b4471 which rendered nd4j arrays not serializable.


nd4j-api/src/main/java/org/nd4j/linalg/api/buffer/BaseDataBuffer.java
```diff
@@ -730,7 +730,7 @@ protected void doReadObject(ObjectInputStream s) {
     protected void doWriteObject(java.io.ObjectOutputStream out) throws IOException {
         out.defaultWriteObject();
         out.writeUTF(allocationMode.name());
-        out.writeLong(length());
+        out.writeInt(length());
```
EronWright: Note that `BaseDataBuffer::length()` is defined as an int.

PR #179, opened 2015-06-24
added parameter server kv store

agibsonccc:
Adding a parameter server for nd4j. Since pistachio won't be deployed to maven central anytime soon we will support a distribution of it (ideally a bit more bundled so we can simplify usage)


agibsonccc:
Migrating everything to c++ for memory management

PR #242, opened 2015-07-27
Add arbitrary-dimension array util functions

AlexDBlack:


AlexDBlack:
Floats.concat(float[]...) and Double.concat(double[]...) only work for 1d and 2d array input. My methods work for everything.

PR #291, opened 2015-08-13
Should not allow out of index

bmsantos:
Problem: This issue was raised while trying to execute the DBNIrisExample in current state of [DL4J examples](https://github.com/deeplearning4j/dl4j-0.0.3.3-examples).

Implemented solution: Moved if condition up in order to guard the intendedIndexes array so that it does not go out of its boundary. 

Note: I'm not really sure if this is even the correct solution. I'm having a hard time understanding what is actually supposed to happen in the method but it seems that there's a smell in the actual implementation since a new "all" array is created to include the allIndexes + new ones in intendedIndexes and then the iteration only happens for the allIndexes. The new ones will never be inserted at the end of the "all" list.


bmsantos:
Introducing a small BDD story could help clarify it. For example:

```gherkin
Feature: ...

Scenario: Should combine intended indexes
Given all indexes
And the intended indexes
When resolve is executed
Then the resulting array length should be the sum of all indexes and new indexed indexes
And it will included the intended indexes
And all other indexes will be reset to its default
```

Issue #306, opened 2015-08-17
Extend INDArray to java.lang.Iterable


whilo:
Hi,

I am currently working on a [core.matrix](https://github.com/mikera/core.matrix) backend for nd4j: https://github.com/whilo/clj-nd4j
It would be beneficial if INDArrays would extend Iterable (to iterate along the first dimension), as then the sequence routines of Clojure would work without wrapping NDArrays. The latter could break direct interoperability of clj-nd4j with deeplearning4j as then new types are introduced and in general would complicate the design. I would also like to avoid that to not accidentally spoil performance.

Is this reasonable? I can provide the implementation if necessary.

Christian

agibsonccc:
Would you want this to be Iterable<INDArray>? Iterable<Double>? Would need more clarity here.

whilo:
You are right, this should be defined very explicitly. If it is iteration along the first dimension it has to return INDArrays, e.g. rows for a matrix. If it is a 1-dimensional vector returning `Iterator<Double>` instead of boxed scalars was nice, but I think this would be inconsistent, right? It also was a problem to have an `Iterator<Object>` in typed JVM languages as explicit casting then becomes necessary. 

It is a bit cumbersome to unbox scalars into doubles, so clatrix, the jblas backend for core.matrix directly returns doubles there so nd-arrays can be treated like arbitrarily nested sequences of doubles. There is also the `clojure.lang.Seqable` interface for which one could explicitly provide this unboxed behaviour, but then nd4j would depend on Clojure.

agibsonccc:
So we have iterators now. Can't link atm because I'm mobile but they do
this kind of nd iteration you'd need.

whilo:
Ok, stupid question: Where? ^^ Note that INDArray (or at least all derived instantiatable classes) need to implement Iterable for a perfect interop. I don't just need some Iterator from somewhere.

PR #374, opened 2015-09-06
Add a numpy-like first axis iterator and implement Iterable for Clojure

whilo:
This is what I meant in #306 and works like numpy + satisfying the `core.matrix` interface. 

(I also stumbled over this documented return value, maybe the docs are just stale, haven't had a deep look into it.)


agibsonccc:
And welcome to clojure!

agibsonccc:
@whilo Good work! I'll be pushing up this release within the next few days to mave central as an rc. This should give you easy access for future development projects.

whilo:
Good, thx!

PR #402, opened 2015-09-17
Newconcat

agibsonccc:


agibsonccc:
Optimization of concat routine.

PR #422, opened 2015-09-22
Fixes bugs in shape resolution or NDArrayIndex

taisukeoe:
In 0.4-rc3.2, some bugs are introduced and this fixes them.
 - NullPointerException at `get` if arguments are NDArrayIndexll.all, due to ShapeOffsetResolution#tryShortCircuit bug
- validate didn't work due to commented out (I'm not sure why it was commented out, though)


agibsonccc:
Optimization pass ;/ Will work on a better solution.

PR #477, opened 2015-10-29
Optimize CPU accum/index accum tasks for nDim=1

AlexDBlack:
***DO NOT MERGE - WIP***


AlexDBlack:
Doesn't generalize to all edge cases. Will replace with something that does.

PR #523, opened 2015-12-03
revert ops back to original design

agibsonccc:
Revert accumulation to do op + update rather than just update


nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/accum/Min.java
```diff
@@ -59,13 +59,23 @@ public String name() {
     }
 
     @Override
+    public float op(float origin, float other) {
+        return FastMath.max(origin, other);
```
AlexDBlack: should be min, not max here

PR #579, opened 2016-02-11
CUDA memory management

raver119:
***WIP; DO NOT CHANGE PLEASE; DO NOT MERGE PLEASE***

Abstracted memory management for CUDA.


raver119:
trick on return values is still to be implemented

raver119:
return values trick stil not implemented

multi-gpu tested for blas, but not for kernels so far.

PR #684, opened 2016-03-17
Disable Google Analytics reporting via HeartBeat.

youhavethewrong:
Most users would likely be surprised to find tracking embedded in a Java library.  Practices like this can very easily blow up in one's face.  If the maintainers of this project would like wide adoption, it would be wise to stop collecting this information.

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/pom.xml
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/pom.xml
@@ -141,20 +141,5 @@
             <version>4.3</version>
         </dependency>
 
-        <dependency>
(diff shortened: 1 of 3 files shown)


agibsonccc:
We're going to need to have a discussion about this. This is a version check for our us to understand what versions of the software are in use and will help us as a company support our users long term.

This also helps us understand where the software is being used so we know where else to allocate resources for translation. We are going to need to start to understand this if we are going to be working with our users long term to support the use cases for our users. If you want to have a real time talk about this, we are very vocal about giving our users access to the devs:
https://gitter.im/deeplearning4j/deeplearning4j

PR #719, opened 2016-03-21
Allocator improvements

raver119:
***WIP; DO NOT MERGE; DO NOT CHANGE***

This PR will cover the following goals:

- Allocator will become abstract. All hardware-dependand logic will be transferred into Mover interface implementations.
- Allocator won't rely on DirectBuffers as original storage, thus removing memory overuse
- Thread->Device affinity will be reconsidered.
- Memory reuse is implemented, to reduce number of global syncs at cuda driver, and time spent for allocation.
- Memory preallocation is implemented, to make cache more effective.

Also, there will be implemented device-only allocation mode for cases, when user knows that he has enough device memory for his tasks.


raver119:
Now it also implements reuse of previously allocated pinned memory.
Cache for device allocations probably shoudn't be implemented, but yet to be decided/prooved.

raver119:
HostAlloc price is somewhere between 50k ns and 500k ns, depending on allocated memory size.
Cache extraction price is somewhere between 130ns and 250ns, depending on moon phase. 

However, exact price of HostAlloc/Free can't be statistically determined, due to hidden device syncrhonization there. So, that price only rises. Memory cache is constant time.

raver119:
Simple preallocation added, looks like it's greatly reducing "warm up" time.

raver119:
Also, host memory garbage collector is reconsidered to become thread-independent.

PR #753, opened 2016-03-29
Expand long indexing to support over Integer.MAX_VALUE elements per NDArray

saudet:
This does not break any builds and does not change the interface of NDArray.

Closes issue deeplearning4j/libnd4j#17.

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java
@@ -1861,7 +1861,7 @@ public abstract class BaseNDArray implements INDArray, Iterable {
         if (ordering() == '\u0000')
             Shape.setOrder(shapeInfo(),Nd4j.order());
 
-        this.length = ArrayUtil.prod(shape);
+        this.length = ArrayUtil.prodLong(shape);
         rank = shape.length;
     }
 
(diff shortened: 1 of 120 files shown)


treo:
@saudet , can you update this PR to cleanly merge again? I'd like to try it out :)

saudet:
Ok, I've updated the branch! Thanks

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java
```diff
@@ -3784,8 +3797,11 @@ public char ordering() {
     @Override
     public int size(int dimension) {
         if (isScalar()) {
+            if(length >= Integer.MAX_VALUE)
```
AlexDBlack: Minor point here: isScalar()==true implies length == 1, so check is unnecessary.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java
```diff
@@ -3810,6 +3826,18 @@ public int rank() {
      */
     @Override
     public int length() {
+        if(length >= Integer.MAX_VALUE)
+            throw new IllegalArgumentException("Length can not be >= Integer.MAX_VALUE");
```
AlexDBlack: Maybe return a different error message here, to avoid confusing users. Length *can* be >= Integer.MAX_VALUE internally, we just can't return it using the length() method.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/accum/StandardDeviation.java
```diff
@@ -96,7 +96,7 @@ public void exec(int... dimension) {
         }
 
         int[] retShape = ArrayUtil.removeIndex(x.shape(), dimension);
-        int nOps = x.tensorssAlongDimension(dimension);
+        long nOps = x.tensorssAlongDimension(dimension);
```
AlexDBlack: Doesn't need to be a long

nyghtowl:
Looks good

PR #769, opened 2016-04-01
createTransform takes extraArgs - used in BaseLayer in dl4j

eraly:


eraly:
Fixed default for relu

PR #904, opened 2016-05-07
Deploop

agibsonccc:
Gets rid of dependency loop


saudet:
If we want the build to work, this merge should be reverted...

PR #905, opened 2016-05-07
Reverdeploop

agibsonccc:
Reverts commenting out the other dependency declaration.


saudet:
Thanks! I think the "Revert" button in GitHub's UI would have done the same thing though

agibsonccc:
There was a lot of other stuff in there. Hence me just adding in your one change.

Issue #917, opened 2016-05-09
Document Additional Dependency if Dev Platform != Deployment Platform


treo:
We have already seen a case where an uberjar created on a Mac didn't work an a Linux machine.

As this case isn't going to be marginal, I think it should be well documented. And maybe even caught on ND4J load, in case that native libraries are present in the classpath but just not for the current platform.

saudet:
I have an additional profile to include all dependencies. But as I mentioned before, users might be better served with additional "pom" artifacts that depend on the java artifacts + one or all of the classifiers. Because, I don't think the way it is now works well or at all for either Gradle or sbt anyway:
https://discuss.gradle.org/t/maven-dependencies-in-profile-section-are-not-supported/2446
http://stackoverflow.com/questions/25961765/sbt-not-resolving-transitive-dependencies-in-the-profile-section-of-pom
In the case of the presets, I don't want to have to create so many pom files for all libraries, but with nd4j we just have 2 right now, so not so bad I guess.

saudet:
I've implemented what I'm talking about above to try out and see: 
https://github.com/deeplearning4j/nd4j/compare/sa_nativebundle 
It could be refined, but there's a couple of issues I've noticed:
1. Maven won't pick up "pom" artifacts by default when listed in dependencies: We have to specify `<type>pom</type>`.
2. At compile time, Maven needs to access to all dependencies even for "pom" artifacts, so unless artifacts for all platforms are made available somewhere, the "-all" artifacts won't install.

PR #920, opened 2016-05-09
Update docs for platform-specific transitive binaries.

crockpotveggies:
Current documentation does not explain to the user how to pull transitive binaries when using a build tool such as Gradle or SBT. This PR addresses that and provides an example of how to specify the targeted build system. It also updates some info for `4.0-rc3.9` regarding the `nd4j-native` group change.

### DIFF
--- a/dependencies.md
+++ b/dependencies.md
@@ -9,6 +9,58 @@ First, you will need to choose the right dependency for your chips:
 
 ![Alt text](../img/backend_table.png) 
 
+## Platform Specific Binaries
+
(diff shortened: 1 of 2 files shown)


crockpotveggies:
I believe this will also help address https://github.com/deeplearning4j/nd4j/issues/917

saudet:
It doesn't address sbt though. Could you try the following patch:
https://github.com/deeplearning4j/nd4j/commit/96c213a9313b5b5edf5ad8fa85276d19c159d4eb
I'm wondering if including, for example, a dependency on `nd4j-native-lib` would work for Gradle.

crockpotveggies:
@saudet I can look into adding an example for SBT, I've had to address platforms in that build tool as well. Will open a new PR. I don't quite understand what that patch does differently? Gradle can't resolve platform profiles

saudet:
Yeah, I think Gradle doesn't pull in "pom" artifacts either, and it's sort of problematic on Maven, so that's not useful I suppose.

PR #957, opened 2016-05-20
Add "org.nd4j.nativeblas.pathsfirst" system property

saudet:
Load BLAS libraries from the system, but only when it is set explicitly by the user, to prevent loading possibly conflicting libraries by default

Would that be satisfactory for now @agibsonccc ?


treo:
From what I've seen this should be satisfactory. There are only a handful of people who want to use something different than what is packaged (including myself), and from those people we can expect that they will either build libnd4j themselves or can use an additional parameter.

PR #1010, opened 2016-06-07
Dataset preprocessor warnings and tests

eraly:
Variance of a constant array was returning nans.
https://github.com/deeplearning4j/libnd4j/pull/233

Dataset preprocessors now has warnings if a roundup is done to avoid div by zero.
More tests for preprocessors.


eraly:
Tests for nans in dataset even if std dev or whatever is zero.

Issue #970, opened 2016-05-22
CUDA State beyond 3.9


raver119:
***Asynchronous execution:***

At this moment, we’re able to provide next-to-100% (or pure 100%) occupancy on any given kernel.
Besides of overall total coolness, that gives us simple fact: this greatly reduces theoretical external parallelism available there. If any single kernel tends to occupy whole device, what’s the point then trying to execute things asynchronously? The only possible gain here is async out-of-order memcpy. Everything else will suffer when executed on full-occupied device.

So, probably async backend should be retired (or left as prototype, for possible future devices, that are better then my gtx 970), and simple delayed execution backend should be introduced, just to make sure that cuda device always has next grid already available for execution.

***Multi-GPU scaling:***

To be discussed separately.

***Shared memory improvements:***

Currently we have pretty small shared memory use, since it’s used for initial Op creation + as reduction buffer for appropriate kernels. And we also want to have low maxrreg count. So since we have average shmem use below 300 bytes, we could be pretty comfortable with some caching involved, probably. However, most of caches do not have anything to cache now: shape & TAD stuff is already at constant memory, and there’s not too much left beyond that. However, we could increase shmem use for local variables, where that could bring performance gains.

***Kernel atomization:***

IndexReduce should be split into Scalar, 1D, 3D and xD kernels, like it was previously done with Reduce kernel.

***Kernels i’m happy with:***

pairwiseStrided
reduceScalar (however, i’d prefer to make it with dp instead of threadfence)
broadcast
ismax filler, however, it maybe worth to make ismax and it’s filler to be the same kernel.

***Kernels i’m NOT happy with:***

All kernels that use ind2sub.
All reduce3 sub-kernels.
Concat HStack for columns would be also nice to have as separate implementation, within HStackGeneric probably.
toFlatten should be executed in the same way as concat does - one kernel call with multiple src pointers.

***FP16 problem:***

We definitely want to have FP16 support. on pascal-generation gpu that should bring some performance boost, together with x2 reduced storage footprint.
The only problem there for us is host memory. JVM should be aware of fp16 somehow, otherwise data use on host side will become impossible. I think the first solution here is JavaCPP fp16 support for host memory. On JVM level methods like getFloat()/getDouble() should just return floats.

raver119:
For asynchronous DataSetIterator prefetch and multi-gpu handling in spark environment following changes are considered:

[...]

PR #1015, opened 2016-06-08
R119 extra args

raver119:
***WIP; DO NOT MERGE;***

- cached extraArgs for ops
- no ops are using extraArgs for write now
- changes for shadow async includes new set of methods for thread-device affinity management
- DELAYED device memory allocation for CUDA, suitable for cluster environment with multiple GPUs, where data originates in one cpu thread, and gets really used in multiple other threads, attached to different devices.
-  Native Dropout Ops introduced, for better performance.

other improvements pending...


raver119:
Manual affinity management as proposed here: https://github.com/deeplearning4j/nd4j/issues/970

raver119:
Delayed device memory allocation for spark-like environments, like proposed here: https://github.com/deeplearning4j/nd4j/issues/970

eraly:
Start Jenkins

Issue #1032, opened 2016-06-20
DataSet (and MultiDataSet) toString issues


AlexDBlack:
DataSet.toString():
- Will throw a NPE if features or labels are null
- Does not display any info about mask arrays

MultiDataSet.toString()
- Doesn't exist (i.e., MultiDataSet does not implement a .toString() method, other than that provided by Object.toString())

eraly:
This is also related to the issue I just closed. Will close this out when I add support the separator.

eraly:
https://github.com/deeplearning4j/nd4j/pull/1042
Added mask info for dataset 
Added toString for multidataset

Issue #930, opened 2016-05-12
Nd4j.writeTxt does not use 'split' arg


AlexDBlack:
https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4j.java#L1684

'split' argument is unused.

eraly:
Closing. 
updates toString for precision and split sep #1036

PR #1036, opened 2016-06-21
updates toString for precision and split sep 

eraly:
toString for precision and split sep 

https://github.com/deeplearning4j/nd4j/issues/930
https://github.com/deeplearning4j/nd4j/issues/1032


eraly:
^^^ This is Jenkins not doing the right thing

eraly:
Start Jenkins

PR #1045, opened 2016-06-23
<WIP> Image preprocessor 

eraly:
https://github.com/deeplearning4j/deeplearning4j/issues/1665


eraly:
You can now scale pixel values of an image to a certain range and give it the number of bits that are used to represent the pixels. 
Added tests as well.

PR #1051, opened 2016-06-24
Introduce artifacts that pull backend binaries for all platforms

saudet:
To use the new functionality users will need to switch dependencies on `nd4j-native` to `nd4j-native-platform`, and `nd4j-cuda-7.5` to `nd4j-cuda-7.5-platform`, but that's it!

It works the same with either Maven, Gradle, or sbt: All binaries for all platforms get downloaded, added to uber JARs (for cross-platform deployment on Spark), etc. But it won't work when building against SNAPSHOT versions where binaries for all platforms are not available. In that case, we need to call Maven with `-Djavacpp.platform=linux-x86_64` or whatever platform we have binaries for. If this still sounds like something we want though, please merge!


agibsonccc:
Could we maybe consider adding a few extra build scripts for this?

saudet:
This doesn't change anything for the build of nd4j or dl4j. It only changes something for the user, and only if they use the new artifacts explicitly, so we don't need additional build scripts.

PR #1054, opened 2016-06-25
Createjavacpppointes

agibsonccc:
Wrap java cpp pointers with data buffers directly. @saudet @raver119


saudet:
It looks like it's now sharing pointers and indexers between buffers. That's alright, but pointers and indexers already have the concept of "offset" and right now it's not being used, so new objects need to be created on each ops call. I thought we could try to figure out how to make buffers use the offset of pointers instead of tracking it separately. This way we could avoid copies on each ops call, but now with these changes, it's going the other way. To make ops calls more efficient we could also pass the offset as an additional parameter along with each pointer, but the whole point of having an offset inside the Pointer class is to avoid passing parameters around like that...

And UByteRawIndexer shouldn't be used explicitly, it won't work when Unsafe isn't available such as on Android.

agibsonccc:
The data buffers have an offset you can use. Feel free to update the PR.

I pinged you with this because the image loading is still slow and the 3
nested for loops approach was a huge bottleneck in me actually getting work
done.

saudet:
Give me some code that's slow and I'll check it out!

agibsonccc:
Biggest change on this branch:
https://github.com/deeplearning4j/Canova/blob/26b3a18aaed7f8d25160cb104031c569b81cb7de/canova-nd4j/canova-nd4j-image/src/main/java/org/canova/image/loader/NativeImageLoader.java#L273-L273

vs:
https://github.com/deeplearning4j/Canova/blob/master/canova-nd4j/canova-nd4j-image/src/main/java/org/canova/image/loader/NativeImageLoader.java#L262

saudet:
I understand what the changes are supposed to fix, but I have not personally experienced that as a bottleneck in any of the examples that I have tried. Could you point me to one of them in the repos that you're having problems with?

PR #1079, opened 2016-07-01
fix for locales

eraly:


eraly:
overriding locales decimal and group separator in tostring

PR #1104, opened 2016-07-09
Fix for views in ndarrays

eraly:


eraly:
@agibsonccc Ahem. You forgot views.

Issue #1050, opened 2016-06-24
Deprecate DataSet normalization methods


AlexDBlack:
I'm proposing that the DataSet normalization methods (normalize(), normalizeZeroMeanZeroUnitVariance()) should be deprecated - they seem to cause a lot of confusion (people not realizing they operate on a single DataSet). Deprecating them, and directing people to use the other/newer normalization functionality should help users to do the right thing (whereas currently it's too easy to do normalization incorrectly).

PR #1109, opened 2016-07-12
Deprecate normalize now that  NormalizerStandardize and NormalizerMinMaxScaler exist

nyghtowl:
Addressing #1050 

@eraly are these methods ready to be deprecated now that you've added  NormalizerStandardize and NormalizerMinMaxScaler?

@AlexDBlack looping you in

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/util/FeatureUtil.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/util/FeatureUtil.java
@@ -58,6 +58,7 @@ public class FeatureUtil {
         return ret;
     }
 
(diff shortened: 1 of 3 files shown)


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/util/FeatureUtil.java
```diff
@@ -58,6 +58,7 @@ public static INDArray toOutcomeMatrix(int[] index, int numOutcomes) {
         return ret;
     }
 
+    @Deprecated
     public static void normalizeMatrix(INDArray toNormalize) {
```
eraly: I think we should keep this one around. I am not familiar with this method.

eraly:
I made a comment inline.

Issue #959, opened 2016-05-20
JVM Crash on TransformOps with equal sized, but differently oriented parameters


treo:
While trying out the example code from https://github.com/deeplearning4j/nd4j/issues/958 I've made a mistake and didn't change the result shape. This resulted in a JVM crash. 

BaseTransformOp already has a check to make sure that both arrays have the same length, but it doesn't check the orientation.

The following code produces reliably a crash with an error dump:
```java
package com.example;

import org.nd4j.linalg.api.ndarray.INDArray;
import org.nd4j.linalg.api.ops.impl.transforms.SoftMax;
import org.nd4j.linalg.api.ops.impl.transforms.SoftMaxDerivative;
import org.nd4j.linalg.factory.Nd4j;

public class App {

    public static void main(String[] args) {
        INDArray input1 = Nd4j.create(1000).transpose();
        INDArray input2 = Nd4j.zerosLike(input1);
        Nd4j.copy(input1, input2);
        INDArray output1 = Nd4j.create(1, 1000);
        Nd4j.getExecutioner().exec(new SoftMax(input1, output1));
        System.out.println("Softmax = " + output1);
        INDArray output2 = Nd4j.create(1,1000);
        Nd4j.getExecutioner().exec(new SoftMaxDerivative(input2, output2));
        System.out.println("Softmax Derivative = " + output2);
    }
}
```

agibsonccc:
Does this affect softmax only?

treo:
With current master it looks like it only happens on SoftMaxDerivative

eraly:
So softmax derivative still crashes on current master. Looking into it.

raver119:
Is CUDA affected?

eraly:
Haven't looked into cuda

raver119:
Ok, i've checked it's not affected. 

sumNumber returns 1.0 for both outputs.

eraly:
@raver119 🎉 
Let me hunt down what it is in cpu.

eraly:
Closing. Merged and fixed in current master. Throws an exception instead of a crash in all such cases.

PR #1116, opened 2016-07-15
Base Transform Op shape check

eraly:
https://github.com/deeplearning4j/nd4j/issues/959

Currently we only check if length is equal if a base transform has two nd arrays as arguments. Mismatched shapes even with the same length can cause JVM crashes. Changed so that it now throws an exception


eraly:
Tests are all passing.

PR #1124, opened 2016-07-19
CachingDataSetIterator

akhodakivskiy:
Caching layer for data set iterator that can speed up DataSet retrieval.


raver119:
Idea is nice, especially InFile datasetcache.

Issue #1128, opened 2016-07-20
Bug with reshape after permute


Jajawah:
When I do a reshape after a permute, I don't get the same result than if I do an inplace permutation.
```
INDArray a = Nd4j.linspace(1, 2, 8).reshape(new int[] { 2, 2, 2 });
INDArray b = a.dup();
int[] rearrange = new int[] { 0, 2, 1 };
a = a.permute(rearrange);
b = b.permutei(rearrange);
System.out.println(a.equals(b));// > return true
int[] newShape = new int[] { 4, 2 };
a = a.reshape(newShape);
b = b.reshape(newShape);
System.out.println(a.equals(b));// >return false :/
```

eraly:
@Jajawah Very good find. Real deal bug. Thank you!

agibsonccc:
https://github.com/deeplearning4j/nd4j/pull/1130

PR #1130, opened 2016-07-21
Forcing ews to -1 for permutei

eraly:
https://github.com/deeplearning4j/nd4j/issues/1128

Doing a permute in place of an array that has ews = 1 just changes the strides and keeps the ews = 1. While the permute operation gives the correct answer downstream operations will give different results when a pairwise transform is invoked, like the reshape in the issue above.


eraly:
And all tests are passing.

PR #1192, opened 2016-08-05
Check for cuBLAS handle creation

raver119:
Check for cuBLAS handle creation


raver119:
We shouldn't have NPE thrown away: https://github.com/deeplearning4j/deeplearning4j/issues/1943

PR #1227, opened 2016-08-20
Update Adam.java

djken2006:
t should start from 1. The first alphat will be NaN if t start from 0.


agibsonccc:
Good find thanks!

PR #1295, opened 2016-09-24
added ability to allow the user to set format of decimal 

claytantor:
I attempted to add a test but when I added the tests they would not get run as part of the test phase. I have tested the implementation using the following test:

```
    @Test
    public void testNd4jArrayString(){

        //2d array
        INDArray arr = Nd4j.create(new float[]{1f,20000000f,40.838383f,3f}, new int[]{2,2});

        //default
        String expected1 = "[[         1.00,20,000,000.00],\n" +
                " [        40.84,         3.00]]";
        String serializedData1 = new NDArrayStrings(",",2).format(arr);
        Assert.assertTrue(serializedData1.equals(expected1));

        //remove commas
        String expected2 = "[[       1.00,20000000.00],\n" +
                " [      40.84,       3.00]]";
        String serializedData2 = new NDArrayStrings(",",2,"######0").format(arr);
        Assert.assertTrue(serializedData2.equals(expected2));

    }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,16 @@ Several of these modules are different backend options for ND4J (including GPUs)
 - scala-notebook = Integration with Scala Notebook
 
 ---
+
+## Building Specific Modules
+
+It is possible to build the project without the native bindings. This can be done
+by specic targeting of the project to build.
+
+```
+mvn clean package test -pl :nd4j-api
+```
+
 ## Documentation
 
 Documentation is available at [nd4j.org](http://nd4j.org/). Access the [JavaDocs](http://nd4j.org/doc/) for more detail.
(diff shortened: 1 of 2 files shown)


agibsonccc:
@claytantor Great eye! Sorry the docs on the nd4j tests are a bit lacking.  We have an nd4j-tests module where you can run tests. We do this for test reuse across nd4j backends. Could you put the test here? https://github.com/deeplearning4j/nd4j/blob/9fe641012eb4bf34a2e7b22e81a85fbe8ee08929/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/api/string/TestFormatting.java#L21-L20 

Thanks!

agibsonccc:
Closed due to lack of activity. Feel free to open again if you move the test over. Thanks!

claytantor:
No worries. I couldn't figure out how to get the tests working. If you want the change feel free to use it from this PR.

PR #1340, opened 2016-10-12
Fix LossKLD

treo:
This was running into NaNs pretty instantly because KL Divergence expects non-zero probabilities.

This changes the score calculation to be exactly the same as it is in Keras (https://github.com/fchollet/keras/blob/master/keras/objectives.py#L51).

I'm not quite sure if the computeGradient method is correct, as is as well. But introducing this change at least results in not NaN scores.


nyghtowl:
Good fix. Will merge in after resolve aeron compilation error on master.

PR #1347, opened 2016-10-19
Add mavent-lint to the build to check for duplicate/redundant deps.

huitseeker:
Avoids issues like #1345

Please consider testing since this removes the hardcoding of a few versions.


huitseeker:
superseded by #1347

PR #1350, opened 2016-10-21
added ability to allow the user to set format of decimal with tests. 

claytantor:
I attempted to add a test but when I added the tests they would not get run as part of the test phase. I have tested the implementation using the following test:

```
    @Test
    public void testNd4jArrayString(){

        //2d array
        INDArray arr = Nd4j.create(new float[]{1f,20000000f,40.838383f,3f}, new int[]{2,2});

        //default
        String expected1 = "[[         1.00,20,000,000.00],\n" +
                " [        40.84,         3.00]]";
        String serializedData1 = new NDArrayStrings(",",2).format(arr);
        Assert.assertTrue(serializedData1.equals(expected1));

        //remove commas
        String expected2 = "[[       1.00,20000000.00],\n" +
                " [      40.84,       3.00]]";
        String serializedData2 = new NDArrayStrings(",",2,"######0").format(arr);
        Assert.assertTrue(serializedData2.equals(expected2));

    }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,16 @@ Several of these modules are different backend options for ND4J (including GPUs)
 - scala-notebook = Integration with Scala Notebook
 
 ---
+
+## Building Specific Modules
+
+It is possible to build the project without the native bindings. This can be done
+by specic targeting of the project to build.
+
+```
+mvn clean package test -pl :nd4j-api
+```
+
 ## Documentation
 
 Documentation is available at [nd4j.org](http://nd4j.org/). Access the [JavaDocs](http://nd4j.org/doc/) for more detail.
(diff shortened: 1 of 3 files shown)


claytantor:
I dont understand why this failed.

claytantor:
If you tell me what I need to do to get this merged I will do it.

AlexDBlack:
Sorry - looks like this just lost in the noise. Feel free to ping us on gitter (or PM me directly) in the future.
Tests failures are probably unrelated - we're seeing some other issues with Jenkins currently.

eraly:
@claytantor The assertion was correct for 'c' order not 'f' order.

PR #1399, opened 2016-11-05
Initial drop of cuda lapack - sample sgetrf

rcorbish:
Added in 1 example of an LAPACK function


rcorbish:
Adding some test cases

PR #1440, opened 2016-11-17
Fix determining column rank when merging MultiDataSets

EdeMeijer:
The indexing into the arrays was in the wrong order, resulting into ArrayIndexOutOfBoundsExceptions when merging MultiDataSets with more than 2 inputs or outputs.

Fixes https://github.com/deeplearning4j/deeplearning4j/issues/2326


AlexDBlack:
Thanks - I was able to reproduce this - almost didn't merge as all existing (and a new test that should have reproduced it, basde on your decription) was passing. Turns out we specifically need num examples < number of arrays.
For future reference, adding a test that catches the condition is a good idea too :)

EdeMeijer:
Yeah, sorry about that. It was the end of a long and frustrating day :p

PR #1444, opened 2016-11-18
Add single file serializer/unserializer for normalization pre-processors

EdeMeijer:
When saving a NormalizerStandardize instance, one had to pass 2 or 4 files depending on whether labels were fit.
With the new MultiNormalizerStandardize, that was 2 * inputs + 2 * outputs, which can become quite unwieldly.

This new utility can serialize and unserialize normalizers to/from single files. As a bonus,
it also saves settings like whether or not to normalize labels.


EdeMeijer:
Did I break something or was the build already falling?

PR #1457, opened 2016-11-24
Remove unnecessary com.sun import

colavitam:
An unnecessary import from com.sun prevents nd4j from building on certain JDKs.


agibsonccc:
Great catch thanks!

PR #1493, opened 2016-12-09
ArrayUtil.java  +=  3D flatten()

gforman44:
Uses System.arraycopy() for speed.  Other routines could adopt this.

RNNs need 3D arrays everywhere for the time dimension.
This helps facilitate such programming.

PR #1494, opened 2016-12-09
Nd4j.java += create() for 3D arrays

gforman44:
RNNs need 3D arrays all the time, so this makes it more convenient, fast, and clearer for new users.

Uses the efficient new flatten in ArrayUtil for 3D arrays.  PL #1493  -- include this FIRST!
Row-major ordering (C).

-gforman44

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4j.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4j.java
@@ -3065,6 +3065,20 @@ public class Nd4j {
         return INSTANCE.create(data, ordering);
     }
 
+    /** Create from 3D array. Fast & handy for Recurrent NN. */
+    public static INDArray create(float[][][] tensor) {
+        float[] flat = ArrayUtil.flatten(tensor);
+        int[] shape = new int[] {tensor.length, tensor[0].length, tensor[0][0].length};
+        return Nd4j.create(flat, shape, 'c');
+    }
+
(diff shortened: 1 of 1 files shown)


raver119:
Please consider adding test for that signature you've added.

raver119:
Also, please consider adding conventional support for F ordering as well.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4j.java
```diff
@@ -3065,6 +3065,20 @@ public static INDArray create(float[][] data, char ordering) {
         return INSTANCE.create(data, ordering);
     }
 
+    /** Create from 3D array. Fast & handy for Recurrent NN. */
+    public static INDArray create(float[][][] tensor) {
+        float[] flat = ArrayUtil.flatten(tensor);
+        int[] shape = new int[] {tensor.length, tensor[0].length, tensor[0][0].length};
+        return Nd4j.create(flat, shape, 'c');
```
raver119: by Nd4j convention, create() signatures use Nd4j.order() if order wasn't specified. And never uses hardcoded 'c' value.

gforman44: I see.  Well, I just thought I'd share my utility extensions, but it looks like a little more time than I have right now to make it general enough to contribute to the shared core.  Perhaps you'd like to include them renamed as createC(), specifying the order.  Someday someone else can generalize them to standard create().

PR #1502, opened 2016-12-13
Implement MultiNormalizerMinMaxScaler + serializer and refactor a lot of related things

EdeMeijer:
I want to use min-max scaling with MultiDataSets but there was no such preprocessor yet, so I created one, and refactored quite a bit to prevent a lot of duplication in code and to clean up a bit. I know it's a bit much in one go, but I didn't really see an easy way to split this up in a logical way. I also found and fixed a few bugs in the old NormalizerMinMaxScaler implementation (using features instead of labels here and there, not properly reverting 3d data), so that's a nice bonus. Let me know if you need additional info or changes.


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/MinMaxStats.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * @param lower row vector of lower bounds
+     * @param upper row vector of upper bounds
+     */
+    public MinMaxStats(@NonNull INDArray lower, @NonNull INDArray upper) {
+        // Check for 0 differences and round up to epsilon
+        INDArray diff = upper.sub(lower);
+        INDArray withEpsilon = Transforms.max(diff, Nd4j.EPS_THRESHOLD);
+        INDArray addedPadding = withEpsilon.sub(diff);
```
agibsonccc: This could be in place since you don't use it anywhere else.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractDataSetNormalizer.java
```diff
@@ -0,0 +1,178 @@
...
+ * @author Ede Meijer
+ */
+@EqualsAndHashCode(callSuper = false)
+abstract class AbstractDataSetNormalizer<S extends NormalizerStats> extends AbstractNormalizer<S> implements DataNormalization {
+    @Setter(AccessLevel.PROTECTED) private S featureStats;
+    @Setter(AccessLevel.PROTECTED) private S labelStats;
+    private boolean fitLabels = false;
+
+    AbstractDataSetNormalizer(NormalizerStrategy<S> strategy) {
+        super(strategy);
```
agibsonccc: Not a big fan of "friendly" for scope. Could you do an explicit protected scope?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractDataSetNormalizer.java
```diff
@@ -0,0 +1,178 @@
...
+    @Override
+    public void fit(DataSet dataSet) {
+        featureStats = (S) newBuilder().addFeatures(dataSet).build();
+        if (isFitLabel()) {
+            labelStats = (S) newBuilder().addLabels(dataSet).build();
+        }
+    }
+
+    S getFeatureStats() {
+        assertIsFit();
```
agibsonccc: Again, same thing friendly -> protected.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractDataSetNormalizer.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+        featureStats = (S) featureNormBuilder.build();
+        if (fitLabels) {
+            labelStats = (S) labelNormBuilder.build();
+        }
+        iterator.reset();
+    }
+
+    abstract protected S.Builder newBuilder();
+
```
agibsonccc: scope first in declarations please

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractMultiDataSetNormalizer.java
```diff
@@ -0,0 +1,243 @@
...
+            result.add((S) builder.build());
+        }
+        return result;
+    }
+
+    private void fitPartial(
+        MultiDataSet dataSet,
+        List<S.Builder> featureStatsBuilders,
+        List<S.Builder> labelStatsBuilders
+    ) {
```
agibsonccc: Move this 1 line up :D. I know this is typical style in c - not in java though (at least not in our code base)

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/MinMaxStrategy.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Normalize a data array
+     *
+     * @param array the data to normalize
+     * @param stats statistics of the data population
+     */
+    @Override
+    public void preProcess(INDArray array, MinMaxStats stats) {
+        if (array.rank() == 2) {
+            array.subiRowVector(stats.getLower());
```
agibsonccc: isMatrix()? Also what about the row vector case? I've ran in to bugs in the past with subiRowVector being applied to row vector. May want to make this explicit.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/StandardizeStrategy.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Normalize a data array
+     *
+     * @param array the data to normalize
+     * @param stats statistics of the data population
+     */
+    @Override
+    public void preProcess(INDArray array, DistributionStats stats) {
+        if (array.rank() == 2) {
+            array.subiRowVector(stats.getMean());
```
agibsonccc: Again same thing here, row vector/column vector explicitly? Make sure vector cases are handled.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/StandardizeStrategy.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Denormalize a data array
+     *
+     * @param array the data to denormalize
+     * @param stats statistics of the data population
+     */
+    @Override
+    public void revert(INDArray array, DistributionStats stats) {
+        if (array.rank() == 2) {
+            array.muliRowVector(filteredStd(stats));
```
agibsonccc: Row/column vector (same comments as above)

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractDataSetNormalizer.java
EdeMeijer: Sure. My IDE was nagging about it, but I think I'd better tell it to shut up.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/MinMaxStats.java
EdeMeijer: Good point.

EdeMeijer: Only the last `sub`, since I use the `diff` twice, but will do.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractMultiDataSetNormalizer.java
EdeMeijer: Do you mean the first parameter? I can't tell which line you mean exactly. Example?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/MinMaxStrategy.java
EdeMeijer: Good point, but I took this straight from the old implementation. I'll fix it though. So what you mean is the input could be a single row of features/labels right?

EdeMeijer: Okay, I added a test and it actually happened to work with row vectors and the broadcast op stuff. However, I changed it to `<= 2` everywhere since that's what was intended and also works.

EdeMeijer:
Okay, I think it's good to go.

PR #1527, opened 2016-12-18
Updated PCA to use SVD.

rcorbish:
Added/updated 4 methods
- pca - return the reduced dimension feature set
- pca_factor - return the factor by which to multiply a feature set to reduce the features

Each has two versions,
- nDims indicate the number of dimensions in the projected feature set
- variance select an amount of variance to keep in the projected feature set

Added a test case


saudet:
Great! Is this ready to merge?

rcorbish:
I can't see what the test failure is - can someone post it please.

saudet:
This one?
```
Error Message

Reconstructed matrix is very different from the original. expected:<0.0> but was:<-0.20078253746032715>

Stacktrace

java.lang.AssertionError: Reconstructed matrix is very different from the original. expected:<0.0> but was:<-0.20078253746032715>
	at org.nd4j.linalg.dimensionalityreduction.TestPCA.testFactorDims(TestPCA.java:49)

Standard Output

o.n.l.BaseNd4jTest - Running org.nd4j.linalg.dimensionalityreduction.TestPCA on backend org.nd4j.linalg.cpu.nativecpu.CpuBackend
o.n.n.Nd4jBlas - Number of threads used for BLAS: 4
o.n.l.BaseNd4jTest - Ending org.nd4j.linalg.dimensionalityreduction.TestPCA
```

saudet:
Seems to be fixed, good to go?

agibsonccc:
I would be fine with merging this if you are.

PR #1539, opened 2016-12-20
javadoc + binary xent scale by two

eraly:


eraly:
@AlexDBlack, please review and make changes as necessary.

Currently we have p,1-p hard coded into the binary cross entropy loss function and which is also accounted for in softmax. So the gradients for softmax+binarycross entropy gets a *2 scaling but works fine for binary cross entropy + sigmoid etc. Doesn't really matter for training as it's a constant scaling but should fix or clarify for users?

Also clarified in javadoc that binary cross entropy only works with one hot encoded labels of size two. And to use negative log likelihood or multiclass cross entropy in other cases.

AlexDBlack:
"Labels are assumed to take one hot values of size two [0,1] or [1,0]"
This is definitely not the case.

BinaryXENT has to support arbitrary sized labels (not restricted to size 2; could be 1, could be N>2); in general it won't be one-hot either.

PR #1547, opened 2016-12-26
Hybrid multi normalizer

EdeMeijer:
This is a new normalizer for MultiDataSets, where the user can specify a different strategy (no normalization, standardization, min-max scaling) for individual inputs and/or outputs. I needed this because I added labels for an embedding layer to my data, which of course shouldn't be normalized, while the other inputs still needed normalization.


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractMultiDataSetNormalizer.java
```diff
@@ -18,7 +17,8 @@
  * @author Ede Meijer
  */
 @EqualsAndHashCode(callSuper = false)
-abstract class AbstractMultiDataSetNormalizer<S extends NormalizerStats> extends AbstractNormalizer<S> implements MultiDataNormalization, Serializable {
+abstract class AbstractMultiDataSetNormalizer<S extends NormalizerStats> extends AbstractNormalizer implements MultiDataNormalization {
+    protected NormalizerStrategy<S> strategy;
```
agibsonccc: If you're going to do "default" could you do protected for the package visibility instead? I like scope being more explicit.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractNormalizer.java
```diff
@@ -1,26 +1,13 @@
...
-    protected AbstractNormalizer() {
-        //
-    }
-
-    protected AbstractNormalizer(@NonNull NormalizerStrategy<S> strategy) {
-        this.strategy = strategy;
-    }
-
+abstract class AbstractNormalizer implements Serializable {
     protected abstract boolean isFit();
```
agibsonccc: What about an interface that extends Normalizer or something similar?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/MinMaxStrategy.java
```diff
@@ -87,8 +90,13 @@ public void revert(INDArray array, INDArray maskArray, MinMaxStats stats) {
...
 
-        if(maskArray != null){
+        if (maskArray != null) {
             DataSetUtil.setMaskedValuesToZero(array, maskArray);
         }
     }
+
+    @Override
+    public NormalizerStats.Builder newStatsBuilder() {
+        return new MinMaxStats.Builder();
```
agibsonccc: Lombok builder?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/StandardizeStrategy.java
```diff
@@ -59,11 +62,16 @@ public void revert(INDArray array, INDArray maskArray, DistributionStats stats)
...
 
-        if(maskArray != null){
+        if (maskArray != null) {
             DataSetUtil.setMaskedValuesToZero(array, maskArray);
         }
     }
 
+    @Override
+    public NormalizerStats.Builder newStatsBuilder() {
+        return new DistributionStats.Builder();
```
agibsonccc: Same thing - lombok?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/serializer/MultiNormalizerHybridSerializer.java
```diff
@@ -0,0 +1,225 @@
...
+        dos.writeDouble(strategy.getMinRange());
+        dos.writeDouble(strategy.getMaxRange());
+    }
+
+    private static NormalizerStrategy readMinMaxStrategy(DataInputStream dis) throws IOException {
+        return new MinMaxStrategy(dis.readDouble(), dis.readDouble());
+    }
+
+    private enum Strategy {
+        NULL,
```
agibsonccc: I usually like being able to use enums for method parmaeters, why not public/protected?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractNormalizer.java
EdeMeijer: Sure, this might be useful to have in a public interface .

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/MinMaxStrategy.java
EdeMeijer: This is not your standard static-factor-method-for-a-new-builder scenario. The normalization strategies have a method for creating a generic *statistics* builder, without having to know the implementation. I notice I should've added some docs here to make this clear though, will update that.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/StandardizeStrategy.java
EdeMeijer: Same

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/serializer/MultiNormalizerHybridSerializer.java
EdeMeijer: On one hand, right now this enum is used exclusively for serde purposes. On the other, an enum for indicating the type of normalization used somewhere in general might be useful, so I could make it public. I'm afraid though that someone might not notice that it's ordinal values are important for serde purposes and just put something in between, effectively breaking all previously saved normalizers. I'd rather keep it here and add a bit fat warning about how to only concatenate new values at the bottom and not touching the other ones.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractMultiDataSetNormalizer.java
EdeMeijer: I'm sorry, don't quite follow you here. Where am I using package visibility rather than protected here? I just moved a field from the parent class one level up to this class.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/MinMaxStrategy.java
agibsonccc: Noted. You may proceed :D. Just make sure to document this.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractNormalizer.java
agibsonccc: Let's wait 0.8.0 here. File an issue mentioning this PR and we will look at a breaking change for 0.8.0. Leave it as is for now.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractMultiDataSetNormalizer.java
agibsonccc: At the abstract class. Could you add an explicit scope before abstract?

EdeMeijer: Ohh sorry yes.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/AbstractNormalizer.java
EdeMeijer: Will do.

EdeMeijer: https://github.com/deeplearning4j/nd4j/issues/1550

PR #1549, opened 2016-12-27
Added gson module to deserialize raw json from the output of NDArray.…

jcastaldoFoodEssentials:
…toString()

### DIFF
--- a/nd4j-serde/nd4j-gson/pom.xml
+++ b/nd4j-serde/nd4j-gson/pom.xml
@@ -11,7 +11,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>nd4j-gson</artifactId>
-    <version>0.7.2-SNAPSHOT</version>
+    <version>0.7.1</version>
     <dependencies>
         <dependency>
             <groupId>org.nd4j</groupId>
--- a/nd4j-serde/nd4j-gson/pom.xml
+++ b/nd4j-serde/nd4j-gson/pom.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>nd4j-serde</artifactId>
+        <groupId>org.nd4j</groupId>
+        <version>0.7.2-SNAPSHOT</version>
+    </parent>
+    <properties>
+        <gson.version>2.8.0</gson.version>
+    </properties>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>nd4j-gson</artifactId>
+    <version>0.7.2-SNAPSHOT</version>
+    <dependencies>
+        <dependency>
+            <groupId>org.nd4j</groupId>
+            <artifactId>nd4j-api</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>${gson.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.nd4j</groupId>
+            <artifactId>nd4j-native</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+</project>
--- a/nd4j-serde/nd4j-gson/pom.xml
+++ b/nd4j-serde/nd4j-gson/pom.xml
@@ -11,7 +11,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>nd4j-gson</artifactId>
-    <version>0.7.1</version>
+    <version>0.7.2-SNAPSHOT</version>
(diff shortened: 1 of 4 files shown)


nd4j-serde/nd4j-gson/src/main/java/org/nd4j/serde/gson/GsonDeserializationUtils.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.List;
+
+public class GsonDeserializationUtils {
+    private static final JsonParser JSON_PARSER = new JsonParser();
+
+    public static INDArray deserializeRawJson(String serializedRawArray) {
+        JsonArray jsonArray = JSON_PARSER.parse(serializedRawArray).getAsJsonArray();
+
+        List<Integer> dimensions = new ArrayList<>();
+        dimensions.add(jsonArray.size());
```
agibsonccc: Mind adding a comment block explaining this? I roughly get what you're doing with the dimensions since the minimum number of dimensions is 2 but would be helpful for others.

nd4j-serde/nd4j-gson/src/main/java/org/nd4j/serde/gson/GsonDeserializationUtils.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        return buildArray(dimensions, serializedRawArray);
+    }
+
+    private static void getSizeMultiDimensionalArray(JsonArray jsonArray, List<Integer> dimensions) {
+        Iterator<JsonElement> iterator = jsonArray.iterator();
+
+        if (iterator.hasNext()) {
+            JsonElement jsonElement = iterator.next();
```
agibsonccc: while?

nd4j-serde/nd4j-gson/src/main/java/org/nd4j/serde/gson/GsonDeserializationUtils.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    private static boolean isArrayWithSingleRow(List<Integer> dimensions) {
+        return dimensions.size() == 1;
+    }
+
+    private static INDArray buildArray(List<Integer> dimensions, String rawArray) {
+        int rank = dimensions.size();
+        int[] shape = new int[rank];
+        for(int i = 0; i < rank; i++) {
```
agibsonccc: Maybe consider using guava here? int[] shape = Ints.toArray(dimensions);

nd4j-serde/nd4j-gson/src/main/java/org/nd4j/serde/gson/GsonDeserializationUtils.java
```diff
@@ -0,0 +1,63 @@
...
+            shape[i] = dimensions.get(i);
+        }
+
+        String[] entries = rawArray.replaceAll("\\[", "").replaceAll("]", "").replaceAll("\\n", "").split(",");
+        double[] entryValues = new double[entries.length];
```
agibsonccc: Any way we could get this in to 1 regex?

nd4j-serde/nd4j-gson/src/test/java/org/nd4j/serde/gson/GsonDeserializationUtilsTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private INDArray buildExpectedArray() {
+        INDArray expectedArray = Nd4j.create(3, 3);
+        expectedArray.putRow(0, Nd4j.create(new double[] { 1, 11, 3 }));
+        expectedArray.putRow(1, Nd4j.create(new double[] { 13, 5, 15 }));
+        expectedArray.putRow(2, Nd4j.create(new double[] { 7, 17, 9 }));
+
+        return expectedArray.reshape(1, 3, 3);
+    }
+}
```
agibsonccc: Add a test for > 3 dimensions? Also what about c ordering vs f ordering?

nd4j-serde/nd4j-gson/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+
+    <artifactId>nd4j-gson</artifactId>
+    <version>0.7.2-SNAPSHOT</version>
+    <dependencies>
+        <dependency>
+            <groupId>org.nd4j</groupId>
+            <artifactId>nd4j-api</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
```
agibsonccc: I'm ok with this being an explicit dependency, but I understand why you did provided (since it'll come with a backend)

nd4j-serde/nd4j-gson/src/main/java/org/nd4j/serde/gson/GsonDeserializationUtils.java
jcastaldoFoodEssentials: I didn't use while because I was operating under the following assumption.
That in an INDArray we can not have a row such as [ 1 , 2, [3, 4] ] and either all elements of an INDArray are either INDArrays themselves or scalars. So if that is the case, then it suffices to only check the first element if it exists to see if it is an array or not. If it is an array, then we must check the first element of that array to see if it's a scalar or array.

agibsonccc: Add a comment block explaining this and I'll be happy :D

nd4j-serde/nd4j-gson/src/test/java/org/nd4j/serde/gson/GsonDeserializationUtilsTest.java
jcastaldoFoodEssentials: I went ahead and added a test for a rank 5 deserialization. What is your vision for features around c and f ordering. By default on my machine Nd4j is defaulting to "c". Should I make my method accept an ordering parameter?

Issue #1553, opened 2016-12-28
Should CachingDataSetIterator delegate the preProcessor?


vjames19:
Currently, the CachingDataSetIterator preProcesses the data every time [next()](https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/iterator/CachingDataSetIterator.java#L145) is called, meaning it will get preProcessed on every epoch.

If the preProcessor is delagated to the sourceIterator, then preProcessing logic can be removed from the CachingDataSetIterator and it would only be calculated once.

Is there a reason for its current behavior? If not, I would love to submit a PR for the change.

raver119:
Yes, that definitely looks like logic flaw in current implementation.

raver119:
If you're going to provide fix for it - please do that, otherwise i'll pick it up.

vjames19:
Yes, I'll link the PR to the issue.

PR #1555, opened 2016-12-28
Delegating the preProcessor on CachingDataSetIterator

vjames19:
#1553


vjames19:
@raver119 Let me know if there is anything else I should do.

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/dataset/CachingDataSetIteratorTest.java
```diff
@@ -56,27 +54,77 @@ public void testInFile() throws IOException {
         FileUtils.deleteDirectory(cacheDir.toFile());
     }
 
+    @Test
+    public void testPreProcessCaching() {
+
```
agibsonccc: Unit test?

vjames19: My bad, that was a transient thought. I just removed it. :)

PR #1557, opened 2016-12-29
Fixed NullPointerException inside Nd4j.rand(1,1) call with jcublas

alexeymateosyan:
**Added check for null on op.x().**

**Before that fix, it crashes with the next stacktrace:**

java.lang.NullPointerException
	at org.nd4j.linalg.api.ops.executioner.DefaultOpExecutioner.checkForCompression(DefaultOpExecutioner.java:64)
	at org.nd4j.linalg.jcublas.ops.executioner.CudaExecutioner.exec(CudaExecutioner.java:2196)
	at org.nd4j.linalg.jcublas.ops.executioner.CudaGridExecutioner.exec(CudaGridExecutioner.java:1032)
	at org.nd4j.linalg.factory.Nd4j.rand(Nd4j.java:2770)

**Call of rand() looks like the next:** 
INDArray S0 = Nd4j.rand(4, 1);


raver119:
Yea, good catch. Caught similar on nd4j-native recently.

PR #1618, opened 2017-02-01
get rid of shape info assertion and put 0 offset back to preserve cac…

agibsonccc:
@raver119


raver119:
I've applied similar fix to ShapeInfoProvider yesterday, with comments why it's there.

PR #1551, opened 2016-12-27
Distributed PS

raver119:
***WIP; DO NOT MERGE;***

Distributed Parameter Server implementation for Word2Vec/ParagraphVectors on Apache Spark.

Only includes ND4j part

PR #1626, opened 2017-02-03
Towards an automated syntax formatting for our projects

huitseeker:
This is an exmaple of usage of the google-code-format tool as a build plugin.
It lets us :

- specify exactly our house style, in contrib/formatter.xml
  For now this initial version is taken from the Google style
  described here https://google.github.io/styleguide/javaguide.html
  implemented here https://github.com/google/styleguide

  In this PR, the file setting this up is [here](https://github.com/huitseeker/nd4j/blob/9d78986d2d47de792f2435155b6173569355b490/contrib/formatter.xml)
  The values can be tested and explored in WYSIWYG fashion in Eclipse, but they're
  pretty sefl-explanatory

- ultimately, we can enforce this style at the build's CI checks. 
  For that, we would just add a commit to this PR that uncomments the following:
  ```xml
              <plugin>
              <groupId>net.revelc.code.formatter</groupId>
              <artifactId>formatter-maven-plugin</artifactId>
              <version>2.0.0</version>
              <!-- <executions> -->
              <!--   <execution> -->
              <!--     <goals> -->
              <!--       <goal>validate</goal> -->
              <!--     </goals> -->
              <!--   </execution> -->
              <!-- </executions> -->
              <configuration>
                <configFile>${eclipseFormatterStyle}</configFile>
              </configuration>
            </plugin>
  ```

- the first commit does not change anything, except that it sets up the plugin
  for action, giving you access to the goals described here:
  http://code.revelc.net/formatter-maven-plugin/plugin-info.html

- Note this formatting can be obtained automatically, either with the Eclipse formatter equipped with the profile mentioned above (in formatter.xml), using the [IntelliJ plugin](https://plugins.jetbrains.com/idea/plugin/8527-google-java-format), or on the CLI, using the [google-java-formatter](https://github.com/google/google-java-format)


agibsonccc:
So I'm in the house formatter guy. Question on this: 
I usually have a few problems with:
spaces: I do int i = 0; i < 5; i++
I see:
i=0;i<5;i++

throughout our codebase and fix it.

I don't do c style braces:
{
}

I do {

}

Does this enforce that? If not how do we configure it?

raver119:
Same story with braces style.

huitseeker:
1. Don't worry about the conflict. The conflicting commit is the reformatting commit, which is tool-generated. It can therefore be re-generated easily.
2. Yes, the Google Java code style enforces Kernighan-Ritchie style braces:
https://google.github.io/styleguide/javaguide.html#s4.1-braces
One example of this enforcement:
https://github.com/deeplearning4j/nd4j/pull/1626/files?w=1#diff-8d2fccb98469a00ac2329f974b6e2155R1548
One exception : empty blocks, which are opened and closed on the same line: 
https://github.com/deeplearning4j/nd4j/pull/1626/files?w=1#diff-1d2730056e767a63c72da5bba6ca95c1R160
See https://google.github.io/styleguide/javaguide.html#s4.1.3-braces-empty-blocks
3. Spaces : yes, they are enforced
https://google.github.io/styleguide/javaguide.html#s4.6.2-horizontal-whitespace
See : https://github.com/deeplearning4j/nd4j/pull/1626/files?w=1#diff-8d2fccb98469a00ac2329f974b6e2155R1573

agibsonccc:
1. Could we put that on .gitignore then?

Beyond that I'm happy merging this.

huitseeker:
You may want to have a look at this PR [without whitespace](https://github.com/blog/967-github-secrets).

Bonus to get all the benefits of this PR:
Also, the further value of this is in having automated checks for formatting, i.e. uncommenting the automatic validation [I talk about above](https://github.com/deeplearning4j/nd4j/pull/1626#issue-205226962). That's only possible if everybody has gone in a sub-project and run a `mvn formatter:format` to convince themselves that when formatting fails their build, it's a one-command fix to make the formatting go and their build pass again.

huitseeker:
And @agibsonccc this is not what I meant by tool-generated — I meant the commit is tool-generated. It touches files that should be under version control.

I think we may want to wait until merging this one that https://github.com/deeplearning4j/nd4j/pull/1551 has gone in — the rebase will be easier for me than @raver119

AlexDBlack:
I'm in favour of this PR, and generally support using the google style guide, and *enforcing* the style guide at build time. I imagine the latter component (enforce style) will take some effort to get the codebase to comply?

Looking through the google code style guide, it's a great starting point, but there's a few things I would want to change:
- Allow wildcard imports: more upsides than downsides IMO https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports
- +4 instead of +2 indentation: I could survive with +2, but we've been using +4 forever, and I see no reason to change https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation
- 120 or 140 instead of 100 character column limit: default in intellij is 120, 100 is too narrow IMO given modern monitors etc. Github seems to be 140 (estimated from looking at code on github , couldn't find a reference for this)  https://google.github.io/styleguide/javaguide.html#s4.4-column-limit

huitseeker:
I changed linewrap, wildcard import and indentation preferences.

agibsonccc:
I can merge this if we get rid of the conflicts.

huitseeker:
Updated /cc @agibsonccc

huitseeker:
Updated /cc @agibsonccc

PR #1659, opened 2017-02-20
Fix build issues to make `compile`, `test` goals perform again on nd4j

### LINKED ISSUE #1658: ND4J compile, test failures because of shade unavailability before package
Issuing 'mvn compile' or 'mvn test' at the root of the nd4j repo fails with messages re:
`package org.nd4j.shade.jackson.databind.annotation does not exist`. Of course, those `org.nd4j.shade.*` packages are not really missing, they are supposed to be created as a result of the `nd4j-shade` project, which uses the [maven-shade plugin](https://maven.apache.org/plugins/maven-shade-plugin/index.html) to shade a jackson dependency. But in phases such as 'compile' and 'test', they aren't created early enough. To wit:
https://asciinema.org/a/2zn7x4kww5rludrq4t27hct3v

On the other hand 'mvn package' works fine.

This is due to the fact that the maven-shade plugin will only produce a shaded jar at the `package` lifecycle build phase, because it expects to have to shade some of the classes generated by the `package`-ing of the project which has shade as a dependency — it's something the shade plugin [enforces in the source of its Mojo](https://github.com/apache/maven-plugins/blob/trunk/maven-shade-plugin/src/main/java/org/apache/maven/plugins/shade/mojo/ShadeMojo.java#L399). 

A possible solution for this is to use the [jetspeed-mvn](https://portals.apache.org/jetspeed-2/buildguide/jetspeed-mvn-plugin.html) plugin in the shaded project, bound to phase `compile`, to create a set of invocations that fall outside of the normal build lifecycle.

huitseeker:
- fix #1658
- comment unused dependency on nd4j-kafka (!?)
- removes more jackson mentions
- removes low-risk spurious files & options


nd4j-serde/nd4j-jackson/src/main/java/org/nd4j/shade/serde/jackson/ndarray/NDArraySerializer.java
```diff
@@ -1,9 +1,9 @@
 package org.nd4j.shade.serde.jackson.ndarray;
 
 
-import com.fasterxml.jackson.core.JsonGenerator;
-import com.fasterxml.jackson.databind.JsonSerializer;
-import com.fasterxml.jackson.databind.SerializerProvider;
+import org.nd4j.shade.jackson.core.JsonGenerator;
+import org.nd4j.shade.jackson.databind.JsonSerializer;
+import org.nd4j.shade.jackson.databind.SerializerProvider;
 import org.nd4j.linalg.api.ndarray.INDArray;
```
agibsonccc: This is wrong. I have 2 versions of this. Check out the shaded package for the nd4j shade version.
I'm explicitly using jackson here.

agibsonccc:
@huitseeker could you look at this again please?

huitseeker:
I removed the jackson removals. How does the last commit look to you, @agibsonccc  ?
https://github.com/deeplearning4j/nd4j/pull/1659/commits/805f5f7f345d6d12d45f5e8d9aff9a195d79559d

huitseeker:
Answering my own question : it's used in `nd4j-serde/nd4j-kryo/src/test/java/org/nd4j/TestNd4jKryoSerialization.java`

huitseeker:
Jackson removals off, commenting commit removed, anything else @agibsonccc @saudet ?

PR #1663, opened 2017-02-24
Maven poms refactor for multiple targets build

syspulse8913:
Modules which should generate mutliple targets have tow pom files. 
Output Example:
```
./nd4j-backends/nd4j-backend-impls/nd4j-cuda-platform/target/nd4j-cuda-8.0-platform-0.7.2.jar
./nd4j-backends/nd4j-backend-impls/nd4j-cuda-platform/target/nd4j-cuda-7.5-platform-0.7.2.jar

./nd4j-serde/nd4j-camel-routes/nd4j-kafka/target/nd4j-kafka_2.11-0.7.2.jar
./nd4j-serde/nd4j-camel-routes/nd4j-kafka/target/nd4j-kafka_2.10-0.7.2.jar
```

Full Build example:
```
cd nd4j; 
mvn clean; mvn install -DskipTests=true -Dmaven.javadoc.skip=true
```
**Limitation**: **mvn clean install** will produce incorrect target!


agibsonccc:
Mind handling the merge conflicts? Thanks!

nd4j-backends/nd4j-backend-impls/nd4j-cuda-platform/pom-cuda75.xml
```diff
@@ -0,0 +1,54 @@
...
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>${nd4j.backend}</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>${nd4j.backend}</artifactId>
+            <version>${project.version}</version>
+            <classifier>${javacpp.platform.linux-x86_64}</classifier>
```
huitseeker: Doesn't this make *all* platforms a dependency of this pom, whereas one of the listed would be sufficient ? Would it be appropriate to add `<optional>true</optional>` to these artefacts ?

nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom-cuda75.xml
```diff
@@ -0,0 +1,307 @@
...
+        <env.LIBND4J_HOME>${basedir}/../../../../libnd4j/</env.LIBND4J_HOME>
+        <buildDirectory>${project.basedir}/target</buildDirectory>
+    </properties>
+
+    <build>
+        <directory>${buildDirectory}</directory>
+        <testSourceDirectory>../../nd4j-tests/src/test/java</testSourceDirectory>
+        <testResources>
+            <testResource>
+                <directory>${project.basedir}/../../nd4j-tests/src/test/resources</directory>
```
huitseeker: I have configured the project with [`directory-maven-plugin`](https://github.com/jdcasey/directory-maven-plugin) to expose the `nd4j` root directory under `$nd4jBaseDir`, perhaps it would be useful to you ?
https://github.com/deeplearning4j/nd4j/blob/master/pom.xml#L408

nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom-cuda75.xml
```diff
@@ -0,0 +1,307 @@
...
+            </testResource>
+        </testResources>
+        <plugins>
+            <plugin>
+                <artifactId>exec-maven-plugin</artifactId>
+                <groupId>org.codehaus.mojo</groupId>
+                <version>1.5.0</version>
+                <executions>
+                    <execution><!-- Run our version calculation script -->
```
huitseeker: It's strange that you need to swap the cuda version link forcibly every recompilation. Isn't `libnd4jcuda.so` linked against a `libcuda.so` of a specific version ? A call to `ldd` would then be a reliable way to see which verison you can expect to work with!
/summoning @raver119 for advice

nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/nativeblas/Nd4jCuda.java
```diff
@@ -50,6 +50,14 @@
 // #endif
 // #include <dll.h>
 
+public static native int tad_threshold(); public static native void tad_threshold(int tad_threshold);
```
huitseeker: Why is this needed ? /cc @agibsonccc @raver119

pom.xml
```diff
@@ -259,6 +254,17 @@
         <module>nd4j-context</module>
         <module>nd4j-backends</module>
         <module>nd4j-parameter-server-parent</module>
+
+
+        <module>nd4j-serde/nd4j-kryo/pom-scala-2_11.xml</module>
+        <module>nd4j-serde/nd4j-camel-routes/nd4j-kafka/pom-scala-2_11.xml</module>
+
+        <module>nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom-cuda75.xml</module>
```
huitseeker: Don't we need a `-cuda80.xml` alongside this ? We don't have a script for cuda version switching.

nd4j-backends/nd4j-backend-impls/nd4j-cuda-platform/pom-cuda75.xml
saudet: Depending on all platforms is what it's supposed to do, that's not an issue.

nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom-cuda75.xml
saudet: Good idea! That's in the original pom.xml file though

nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom-cuda75.xml
saudet: I'm not sure I understand why we need to do that...

nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/nativeblas/Nd4jCuda.java
saudet: Probably a failed merge @syspulse8913 ?

saudet:
@syspulse8913 Sorry it took me a bit of time to get back to this. So, it looks like there is a lot extra stuff in this PR that is unrelated to the changes you actually want to bring in. Are they caused by a failed merge? Could you confirm and fix if necessary? If I understand correctly, this PR is about adding cuda75.sh and cuda80.sh plus the related logic in pom.xml files, etc. If so, I'm not sure I understand the need for all this because the compiler as invoked by JavaCPP will look into both libnd4j/blasbuild/cuda-7.5 or libnd4j/blasbuild/cuda-8.0 before looking into libnd4j/blasbuild/cuda. We don't actually need libnd4j/blasbuild/cuda, it's just there as a tag for Maven to through an error message. It can also be as a fallback when the user tries some unknown version of CUDA, but it **should not** be used for known and supported versions like 7.5 or 8.0.

saudet:
@syspulse8913 Sorry, went through the files too fast. Ok, I see there are a bunch of pom.xml files ending with "scala-2_11.xml", but then, in parent pom.xml files, we have lines like these:
```
<module>nd4j-serde/nd4j-kryo/pom-scala-2_11.xml</module>
```
I thought the advantage of such a scheme is that we wouldn't need to run scripts to change the versions of Scala, Spark, CUDA, etc. But it looks like we still need to run scripts to change  lines like above, so I am not sure I understand the advantage. Could you explain? Thanks!

pom.xml
saudet: Actually, no, I think I get it now. The default pom.xml file does 8.0, so we only need 7.5 as additional. The same for the Scala versions. The default pom.xml files still do 2.10 in this branch. That's what confused me.

Seems to be a pretty good idea actually and (after removing cuda75.sh and cuda80.sh) it works for me! Maven seems to build everything fine in one shot. What do you think @AlexDBlack?

agibsonccc:
I'm going to close this due to inactivity if I don't see a reply in the next few days here. I want to make sure this is still being worked on.

saudet:
It's being worked on, but we're not going to merge this before the release. We'll do this kind of big change after the release.

agibsonccc:
Closing due to inactivity.

PR #1670, opened 2017-02-27
Variosu fixes

agibsonccc:
Param server fixes.


gagatust:
`list.forEach((message) -> {`

Java8?

raver119:
Nope, only in selected modules like this one. It relies on Aeron, which is java8.  More like exception from the general java7 rule.

raver119:
I mean - we're 100% sure param server implementations will never be used on Android. So we don't care about android requirements here.

gagatust:
OK

PR #1673, opened 2017-02-28
Fix for "cannot convert int[] to int" compile-time error

turambar:
Fix for "cannot convert int[] to int" compile-time error occurring inside `d4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/Nd4jTestsC.java`.

Problem was line 1939:

```INDArray repeated = x.repeat(new int[] {2});```

which should be

```INDArray repeated = x.repeat(1, new int[] {2});```

FTFY.


turambar:
Already fixed on master.

Issue #1550, opened 2016-12-27
Consider cleaning up preprocessor/normalizer interfaces (breaking changes)


EdeMeijer:
As discussed in https://github.com/deeplearning4j/nd4j/pull/1547, with the introduction of normalization for `MultiDataSet`, the interface ecosystem has become a bit messy. There's quite some duplication between the interfaces, but unifying them would incur some breaking changes. We could consider this for 0.8.0.

The interface structure right now:
```
DataSetPreProcessor -> DataNormalization

MultiDataSetPreProcessor -> MultiDataNormalization
```

With the use of generics, these could be (partially) merged. Also, we could consider making method names more consistent here and there and removing redundant methods.

agibsonccc:
Next release is 0.8.0 so this might be a good pull request.

EdeMeijer:
How much time do I have? :P

agibsonccc:
We are doing code freeze in about a week or so. If you want to get this in great if you can't it's not an issue.

EdeMeijer:
Alright, I'll see what I can do. Sounds like it'll be right in time to get 0.8.0 in my production stack I'm building right now, so this would be nice to have for myself too.

agibsonccc:
@EdeMeijer updates here?

EdeMeijer:
Sort of. I poked around but soon realized it's not as straightforward as I thought it would be. The biggest issue is that DataSet normalizers work with a single `INDArray` for features and labels while MultiDataSet normalizers work with `INDArray[]` arrays everywhere. This is not something that can be easily generalized with generics.

However, if there's still time, there's something else I'd like to do. I reworked the whole normalizer serde stuff to support all multi normalizer types and package everything in single files, but I realized it would be better to encode the normalizer's type in the serialized data, so it can be restored without having to know the type beforehand. This way, we could go back from a `MultiNormalizerStandardizeSerializer`, `MultiNormalizerMinMaxScalerSerializer` etc to just a `NormalizerSerializer` facade that works transparently for all types. Also, I'm missing a `restore(InputStream)` (only implemented `restore(File)`).

I think I can finish this in a few hours and should be a fairly safe change.

huitseeker:
See also #1425 #1650

raver119:
@AlexDBlack @eraly  is this still an issue? I remember there were normalizer-related changed some time ago...

AlexDBlack:
Yeah, there was a PR that improved functionality for normalizers a long time ago.

PR #1674, opened 2017-02-28
Unify data normalizer serialization functionality

EdeMeijer:
Follow up for https://github.com/deeplearning4j/nd4j/issues/1550

This PR introduces a single entry point for serializing and unserializing data normalizers transparently. To support both DataSet and MultiDataSet normalizers, I introduced a base interface for both types of normalizers. Users can also define custom normalizers and serialization strategies and register them with the default serializer.


EdeMeijer:
Any thoughts?

PR #1710, opened 2017-03-17
This is a Intropro example

smaryn:
NOTE: Do not merge those changes
Added different profiles for serving different repository integrated with Jenkins


huitseeker:
@smaryn Is this still needed ?

PR #1711, opened 2017-03-17
[WIP] Sparse implementation

audreyloeffel:


huitseeker:
Ping @NonaryR.

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/CpuSparseMatrixFactory.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @author Audrey Loeffel
+ */
+public class CpuSparseMatrixFactory extends ISparseMatrixFactory {
+
+    public CpuSparseMatrixFactory(){}
+
+    static{
+        Nd4j.getBlasWrapper();
```
raver119: Are you 100% sure here? There's lots of static stuff happens at nd4j initialization, and it's quite easy to get into race condition here. Like 0.7.2 issue with CudaEnvironment bean

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/blas/SparseCpuLevel1.java
```diff
@@ -0,0 +1,14 @@
...
+import org.nd4j.linalg.api.blas.impl.SparseBaseLevel1;
+import org.nd4j.linalg.factory.Nd4j;
+import org.nd4j.nativeblas.Nd4jBlas;
+
+/**
+ * @author Audrey Loeffels
+ */
+public class SparseCpuLevel1 extends SparseBaseLevel1 {
+
+    private Nd4jBlas nd4jBlas = (Nd4jBlas) Nd4j.factory().blas();
```
raver119: I'm not sure if native sparse provider will be in the same class dense blas is. Maybe better introduce separate object? We can route op calls between them from java side.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/ISparseMatrix.java
```diff
@@ -0,0 +1,65 @@
...
+import org.nd4j.linalg.api.buffer.FloatBuffer;
+import org.nd4j.linalg.api.buffer.IntBuffer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Created by audrey on 3/2/17.
+ */
+public abstract class ISparseMatrix /*implements INDArray*/ {
+    private static final double ALLOCATION_RATION = 0.3;
```
raver119: Let's make this option configurable. I.e. via backend options. For CUDA it's not our problem, sure, but for CPU it will be our responibility to reallocate.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/impl/BaseLevel.java
```diff
@@ -12,9 +12,7 @@
 public abstract class BaseLevel {
 
 
-    public float[] getFloatData(INDArray buf) {
-        return BlasBufferUtil.getFloatData(buf);
-    }
+    public float[] getFloatData(INDArray buf) { return BlasBufferUtil.getFloatData(buf); }
```
raver119: Are we actually using this method anywhere? It should be legacy remnants...

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/impl/SparseBaseLevel.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.nd4j.linalg.api.buffer.DataBuffer;
+import org.nd4j.linalg.api.ndarray.INDArray;
+
+/**
+ * @author Audrey Loeffel
+ */
+public abstract class SparseBaseLevel {
+    //TODO implementations
+    public float[] getFloatData(INDArray buf){
```
raver119: Once again, i'm in doubt do we really need these methods. It's a legacy stuff, we could probably just remove them from interface, since they have 0 value. to get double[] from INDArray or DataBuffer all you need is 1 line:

```
buf.data().asDouble();
```

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/ISparseMatrix.java
raver119: And it should not be coefficient, reallocation step is assumed here.

I.e. initially you've allocated 100 bytes. Then you get used 99 of them, and need some more space. You should reallocate, but what amount of ram should be requested? Usually step is x2 of the current allocation, and that's what we should make configurable - step size.

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/SparseMatrixTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+        assert(expectedValues == values);
+        assert(expectedColumns == columns);
+        assert(expectedPointerB == pointerB);
+        assert(expectedPointerE == pointerE);
+        assert(expectedShape == shape);
+    }
+
+    @Test
+    public void shouldReplaceValueAtAGivenPosition(){
```
huitseeker: Nice 👍

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseSparseNDArray.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+@Slf4j
+public class BaseSparseNDArray implements ISparseNDArray {
+
+    protected static final double THRESHOLD_MEMORY_ALLOCATION = 0.5;
+
+    protected transient volatile long nnz = -1;
+    protected int nbRows, nbColumns;
+    protected Boolean isVector = null;
+    protected Boolean isMatrix = null;
```
agibsonccc: Use primitive types rather than objects where you can.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseSparseNDArray.java
```diff
@@ -0,0 +1,39 @@
...
+    protected Boolean isScalar = null;
+
+    protected DataBuffer reallocate(DataBuffer buffer) {
+        int newSize = (int) buffer.length() * 2; // should be bound to max(nnz, size*2)
+        DataBuffer newBuffer = null;
+        if (buffer instanceof DoubleBuffer){
+            newBuffer = new DoubleBuffer(newSize);
```
agibsonccc: You should be using Nd4j.createBuffer not instantiating a new buffer yourself.

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/CpuCSRSparseNDArray.java
```diff
@@ -0,0 +1,141 @@
...
+ * @param columns Element i of the integer array columns is the number of the column in A that contains the i-th value
+ *                in the values array.
+ * @param pointerB Element j of this integer array gives the index of the element in the values array that is first
+ *                 non-zero element in a row j of A. Note that this index is equal to pointerB(j) - pointerB(1)+1 .
+ * @param pointerE An integer array that contains row indices, such that pointerE(j)-pointerB(1) is the index of the
+ *                 element in the values array that is last non-zero element in a row j of A.
+ * @param shape Shape of the matrix A
+*/
+    public CpuCSRSparseNDArray(double[] data, int[] columns, int[] pointerB, int[] pointerE, int[] shape) {
+        assert(data.length == columns.length);
```
agibsonccc: Don't use assertions.  Use guava's PreConditions instead.

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/CpuCSRSparseNDArray.java
```diff
@@ -0,0 +1,141 @@
...
+        this.pointerB = new IntBuffer(pointersSpace);
+        this.pointerB.setData(pointerB);
+        this.pointerE = new IntBuffer(pointersSpace);
+        this.pointerE.setData(pointerE);
+    }
+
+    public ISparseNDArray putScalar(int row, int col, double value){
+        // TODO use shape information to get the corresponding index ?
+
+        assert(row < nbRows);
```
agibsonccc: Same thing: Don't use assert. Use Preconditions from guava.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseSparseNDArray.java
audreyloeffel: I did the same as https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java#L99-L101
Should I also change these lines and use primitive boolean or is there a reason for using object in this case?

agibsonccc: Ahh good catch thanks. We used object because of the lazy init. Carry on then :D.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/impl/BaseLevel.java
agibsonccc: I don't think so.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/impl/BaseLevel1.java
```diff
@@ -25,17 +25,26 @@
...
+     * @return the vector-vector dot product of X and Y
      */
     @Override
     public double dot(int n, double alpha, INDArray X, INDArray Y) {
         if (Nd4j.getExecutioner().getProfilingMode() == OpExecutioner.ProfilingMode.ALL)
             OpProfiler.getInstance().processBlasCall(false, X, Y);
 
+        if (X.isSparse() && !Y.isSparse()){
+            return Nd4j.getSparseBlasWrapper().level1().dot(n, alpha, X, Y);
+        } else if (!X.isSparse() && Y.isSparse()) {
```
agibsonccc: Mind explaining the difference here? Not obvious from looking at the code here.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/impl/BaseLevel1.java
```diff
@@ -300,6 +324,9 @@ public void copy(INDArray x, INDArray y) {
         if (Nd4j.getExecutioner().getProfilingMode() == OpExecutioner.ProfilingMode.ALL)
             OpProfiler.getInstance().processBlasCall(false, x, y);
 
+        if(x.isSparse() || y.isSparse()){
+            Nd4j.getSparseBlasWrapper().level1().copy(x, y);
+        }
         if (x.data().dataType() == DataBuffer.Type.DOUBLE) {
             DefaultOpExecutioner.validateDataType(DataBuffer.Type.DOUBLE, x, y);
```
agibsonccc: @raver119 I feel like this should be in a util class of some kind?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/impl/SparseBaseLevel2.java
```diff
@@ -0,0 +1,131 @@
...
+
+import org.nd4j.linalg.api.blas.Level2;
+import org.nd4j.linalg.api.complex.IComplexNDArray;
+import org.nd4j.linalg.api.complex.IComplexNumber;
+import org.nd4j.linalg.api.ndarray.INDArray;
+
+/**
+ * @author Audrey Loeffel
+ */
+public abstract class SparseBaseLevel2 extends SparseBaseLevel implements Level2{
```
agibsonccc: Remember to run the formatter when you commit.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java
```diff
@@ -2012,6 +2013,10 @@ public boolean isView() {
          */
         return Shape.offset(shapeInformation) > 0 || length() < data().length() || data().originalDataBuffer() != null;
     }
+    @Override
+    public boolean isSparse(){
+        return isSparse;
```
agibsonccc: Couldn't this just return false? You have a sparse ndarray as well.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseSparseNDArrayCSR.java
```diff
@@ -0,0 +1,159 @@
...
+        checkArgument(pointerB.length == pointerE.length);
+
+        if (shape.length == 2) {
+            nbRows = shape[0];
+            nbColumns = shape[1];
+        } else if (shape.length == 1) {
+            nbRows = 1;
+            nbColumns = shape[0];
+        } else {
+            // ???
```
agibsonccc: Don't worry about this.  rows/columns is a short cut for matrix specific operations. Look at the BaseNDArray if you have a question on the implementation there.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/BaseBlasWrapper.java
```diff
@@ -21,6 +21,10 @@
 
 public abstract class BaseBlasWrapper implements BlasWrapper {
 
+    public BaseBlasWrapper() {
```
agibsonccc: This feels redundant. This is implicit in java.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/BaseSparseBlasWrapper.java
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    @Override
+    public IComplexNumber dotc(IComplexNDArray x, IComplexNDArray y) {
+        return null;
+    }
+
+    @Override
+    public IComplexNumber dotu(IComplexNDArray x, IComplexNDArray y) {
+        return null;
```
agibsonccc: Don't worry about implementing complex right now. Just do: throw new UnsupportedOperationException() on these operations.

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/blas/SparseCpuBlas.java
```diff
@@ -0,0 +1,126 @@
...
+ */
+public class SparseCpuBlas extends SparseNd4jBlas {
+
+    /**
+     * Converts a character
+     * to its proper enum
+     * for row (c) or column (f) ordering
+     * default is row major
+     */
+    static int convertOrder(int from) {
```
agibsonccc: I feel like these could be in a combined util class with the dense version?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/impl/BaseLevel1.java
audreyloeffel: MKL doesn't support dot if both vectors are sparse. As Raver commented previously, I need to swap the operands depending on which vector is sparse. Is it better to check whether there is at least one sparse vector here and then swap if needed  in SparseBaseLevel (or throw an exception if both sparse) ?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/impl/BaseLevel.java
audreyloeffel: So should I delete them ? Does this class has a reason to exist without those methods? It seems to be not used

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/blas/SparseCpuBlas.java
audreyloeffel: It doesn't change anything whether the values come from MKL or openblas ?

agibsonccc: Feeling like what we could do is code share the conversions: https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/blas/CpuBlas.java

Then just convert the int to the proper enum?

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/blas/SparseCpuLevel1.java
```diff
@@ -0,0 +1,271 @@
...
+     * @param N The number of elements in vector X
+     * @param alpha
+     * @param X a sparse vector
+     * @param pointers A DataBuffer that specifies the indices for the elements of x.
+     * @param Y a dense vector
+     *
+     * */
+    @Override
+    protected void daxpyi(int N, double alpha, INDArray X, DataBuffer pointers, INDArray Y){
+        cblas_daxpyi(N, alpha, (DoublePointer) X.data().addressPointer(), (IntPointer) pointers.addressPointer(),
```
agibsonccc: @huitseeker do we have any rules in the formatter for excessively long lines?
I usually truncate those with:
longMethodCall(
arg1,
arg2,..
);

@audreyloeffel if we don't have it yet we will add that and then I'lll ask you to run the formatter here.

nd4j-backends/nd4j-api-parent/nd4j-native-api/src/main/java/org/nd4j/nativeblas/SparseNd4jBlas.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.nd4j.linalg.api.blas.Blas;
+
+public abstract class SparseNd4jBlas implements Blas {
+
+    public SparseNd4jBlas(){
+
+    }
+
+    @Override
```
agibsonccc: Not sure you need the overrides here for the abstract class?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/indexing/SparseCSRIndex.java
```diff
@@ -0,0 +1,7 @@
+package org.nd4j.linalg.indexing;
+
+/**
+ * @author Audrey Loeffel
+ */
+public class SparseCSRIndex /*extends INDArrayIndex */{
```
agibsonccc: Got a reason for not using the existing indexing? Are you going to need additional information?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/BaseSparseNDArrayFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import org.nd4j.linalg.api.blas.*;
+
+/**
+ * @author Audrey Loeffel
+ */
+public abstract class BaseSparseNDArrayFactory implements SparseNDArrayFactory {
+    protected char order; // c or f
+    protected Blas blas;
+    protected Level1 level1;
+    protected Level2 level2;
```
agibsonccc: Any reason we couldn't use the existing NDArrayFactory and just use a createSparse method?

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/blas/SparseCpuLevel1.java
audreyloeffel: I did run `mvn formatter:format` before committing. So this might not be included. I'll run it again once the formatter will be fixed.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/BaseSparseNDArrayFactory.java
audreyloeffel: I'm not sure about that. I cannot use the same implementation of the factory (CpuNdArrayFactory) because they are using different blas. Using NDArrayFactory instead of SparseNdArrayFactory would ask to implement all the dense methods in the sparse instance and there will be a lot of unsupported methods. Does it make sense to do it anyway?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/indexing/SparseCSRIndex.java
audreyloeffel: Nope, it should've been removed earlier...

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/BaseSparseNDArrayFactory.java
agibsonccc: @saudet What are your thoughts here? I know the people who consume the dl4j backend are going to expect sparse and dense to be usable together even if it's less than ideal. How would this affect the presets?

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/blas/SparseCpuLevel1.java
agibsonccc: Thanks!

nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/linalg/jcublas/blas/JcusparseLevel1.java
```diff
@@ -1,9 +1,131 @@
...
 
 import org.nd4j.linalg.api.blas.impl.SparseBaseLevel1;
+import org.nd4j.linalg.api.buffer.DataBuffer;
+import org.nd4j.linalg.api.ndarray.INDArray;
 
 /**
  * @author Audrey Loeffel
  */
 public class JcusparseLevel1 extends SparseBaseLevel1 {
+
```
agibsonccc: 1 suggestion here: study the current lapack implementation to look at how to interact with databuffers and cuda pointers.

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/nativeblas/Nd4jCpu.java
```diff
@@ -4577,45 +4577,6 @@ public native void execMetaPredicateReduceFloat(@Cast("Nd4jPointer*") PointerPoi
 
     public native @Cast("Nd4jPointer") Pointer pointerForAddress(long address);
```
agibsonccc: @audreyloeffel this diff is wrong. Can you double check what you did against current master? https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/nativeblas/Nd4jCpu.java

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/blas/SparseCpuBlas.java
saudet: @audreyloeffel ND4J should allocate the data, so it doesn't matter what processes it

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/BaseSparseNDArrayFactory.java
saudet: Maybe make SparseNdArrayFactory, etc a subclass of the existing concrete classes for dense matrices? I don't think MKL forces us to allocate memory in a special way. In any case, presets aren't affected in any way.

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/SparseNDArrayTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Test
+    public void shouldCreateSparseMatrix() {
+        INDArray matrix = Nd4j.createSparseCSR(values, columns, pointerB, pointerE, shape);
+        //TODO
+    }
+
+    @Test
+    public void shouldAddValueAtAGivenPosition(){
+        /*
```
agibsonccc: Just going to note for indexes here, you're going to need to redo the indexing engine at some point for sparse. The offsets and strides it uses are for dense only.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseSparseNDArrayCSR.java
```diff
@@ -213,16 +213,19 @@ public boolean isMatrix() {
...
 
     @Override
     public int[] shape() {
+        //TODO
         return super.shape();
     }
 
     @Override
     public int[] stride() {
+        // Sparse matrix hasn't any stride. return .. ?
```
agibsonccc: Hmm maybe look at how numpy does it..I'd imagine there's gota be a stride concept of some kind?

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/CpuSparseNDArrayFactory.java
```diff
@@ -25,6 +25,63 @@
...
+    }
+
+    @Override
+    public INDArray create(float[] data, int[] shape, int[] stride, int offset) {
+        return null;
+    }
+
+    @Override
+    public IComplexNDArray createComplex(double[] data, int[] shape, int[] stride, int offset) {
+        return null;
```
agibsonccc: Same thing you can throw new UnsupportedoperationExample on the sparse stuff for now.

The create operations for things like the float arrays I'd imagine you could do a toSparse() maybe?

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/CpuNDArrayFactory.java
```diff
@@ -1155,9 +1155,15 @@ public INDArray createFromNpyFile(File file) {
         System.out.println(otherBytes);
         ByteBuffer directBuffer = ByteBuffer.allocateDirect(pathBytes.length).order(ByteOrder.nativeOrder());
         directBuffer.put(pathBytes);
+        directBuffer.put(pathBytes);
```
agibsonccc: Any reason we're doing this twice?

nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml
```diff
@@ -39,6 +39,17 @@
             <version>${openblas.version}-${javacpp-presets.version}</version>
             <classifier>${dependency.platform}</classifier>
         </dependency>
+        <dependency>
+            <groupId>org.bytedeco.javacpp-presets</groupId>
+            <artifactId>mkl</artifactId>
```
agibsonccc: @saudet Are we really going to have 2 deps for this? The presets are pretty heavy.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/impl/BaseLevel1.java
```diff
@@ -123,6 +139,10 @@ public IComplexNumber nrm2(IComplexNDArray arr) {
      */
     @Override
     public double asum(INDArray arr) {
+
+        if (arr.isSparse()){
+            return Nd4j.getSparseBlasWrapper().level1().asum(arr);
```
agibsonccc: I like this direction.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/impl/BaseLevel1.java
```diff
@@ -268,6 +295,10 @@ public void swap(INDArray x, INDArray y) {
         if (Nd4j.getExecutioner().getProfilingMode() == OpExecutioner.ProfilingMode.ALL)
             OpProfiler.getInstance().processBlasCall(false, x, y);
 
+        if(x.isSparse() || y.isSparse()) {
+            Nd4j.getSparseBlasWrapper().level1().swap(x, y);
```
agibsonccc: Needs a return statement?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/impl/BaseLevel1.java
agibsonccc: /ping @raver119

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/SparseFormat.java
```diff
@@ -0,0 +1,8 @@
+package org.nd4j.linalg.api.ndarray;
+
+/**
+ * @author Audrey Loeffel
+ */
+public enum SparseFormat {
+    CSR /* CSC, COO, ...*/
```
agibsonccc: May want a little more docs on the individual formats here.

nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/linalg/cpu/nativecpu/CpuSparseNDArrayFactory.java
```diff
@@ -0,0 +1,388 @@
...
+    }
+
+    @Override
+    public IComplexNDArray createComplex(INDArray arr) {
+        return null;
+    }
+
+    @Override
+    public IComplexNDArray createComplex(IComplexNumber[] data, int[] shape) {
```
agibsonccc: Same thing as the rest of the complex. Add throw new UnsupportedOperationException()

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseSparseNDArrayCSR.java
```diff
@@ -311,40 +312,22 @@ public INDArray subArray(ShapeOffsetResolution resolution) {
...
+                offsets[0] = (int) resolution.getOffset() / shape()[1];
+                offsets[1] = (int) resolution.getOffset() % shape()[1];
             }
-            //System.out.println("from row " + firstRow + " to row " + lastRow);
-            //System.out.println("from col " + firstElement + " to col " + lastElement);
+            int firstRow = offsets [0];
+            int lastRow = firstRow + shape[0];
+            int firstElement = offsets [1];
+            int lastElement = firstElement + shape[1];
+
```
agibsonccc: So a few things - shape() should be replaced with size(index). Also - how much are you looking at actual sparse ndarrays? I would like us to focus on the general case here. We should be able to expand this logic out for ndarrays not just matrices. Matrices should be a special case of general ndarray logic.

PR #1716, opened 2017-03-20
Intropro081

smaryn:
Updated POM file with profiles for different repositories.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,19 +34,6 @@
         <maven>3.0</maven>
     </prerequisites>
 
-<!--    <distributionManagement>
-        <snapshotRepository>
-            <id>sonatype-nexus-snapshots</id>
-            <name>Sonatype Nexus snapshot repository</name>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -34,7 +34,7 @@
         <maven>3.0</maven>
     </prerequisites>
 
-    <distributionManagement>
+<!--    <distributionManagement>
```
huitseeker: This information can be found in the sonatype-nexus plugin, so that I think you can remove these lines.

saudet: I'd actually like to keep the old stuff around for a bit, even if in comments, just in case I need it to do another release with it. Would that be OK?

huitseeker: AFAICT, the `sonatype-nexus` profile reflects very precisely the information that is in comments. So that information doesn't disappear, I believe you can do a release with it still.

saudet: Right, and the same would be true for the other commented out lines below. Ok, I would be fine just removing all of them, these and the ones in the other repositories. @smaryn Sounds good?

PR #1727, opened 2017-03-27
Fix bugs in ScalarMin

gagatust:
This class has "copy-paste  error".
Also, I think ScalarMin.java  and all classes in org.nd4j.linalg.api.ops.impl.scalar package have bad implementation.
For examples:
```java
    @Override
    //I see the interface and think the other parameter must be used, but it is ignored
    public IComplexNumber op(IComplexNumber origin, double other) { 
        if (origin.absoluteValue().doubleValue() < complexNumber.absoluteValue().doubleValue())
            return origin;
        return complexNumber; 
    }

    @Override
    //the other is ignored
    public double op(double origin, double other) { 
        //it will be better to throw UnsupportedOperationException ("use op(double) method");
        return FastMath.min(origin, num.doubleValue()); 
    }

    @Override
    //good implementation
    public double op(double origin) {
        return FastMath.min(origin, num.doubleValue());
    }
```

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/scalar/ScalarMin.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/scalar/ScalarMin.java
@@ -58,55 +58,55 @@ public class ScalarMin extends BaseScalarOp {
 
     @Override
     public String name() {
-        return "max_scalar";
+        return "min_scalar";
     }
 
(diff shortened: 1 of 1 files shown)


agibsonccc:
@gagatust we haven't actually been using the java based operations for a long time now. That functionality got moved to c. Thanks for noticing though!

gagatust:
@agibsonccc do you plan to delete or mark "deprecated" this package?

raver119:
It's not about package, ops are still used and will be used in future. It's just math happens outside of java. And your fix only affects java code, which isn't used at this moment.

PR #1745, opened 2017-04-02
Adds numpy serialization and interop for nd4j

agibsonccc:

Adds java part of numpy serialization.

## How was this patch tested?
1 unit test - needs more yet.

Please review


saudet:
`*RawIndexer` shouldn't be imported because those won't be usable on Android and other JVMs where `Unsafe` isn't available. Why do you need them explicitly?

agibsonccc:
I needed an indexer for creating a data buffer with nd4j. If you have an alternative I'm open to it.

saudet:
Indexers are fine, but use say `IntIndexer.create(pointer)` without downcast.

agibsonccc:
k added

Issue #1224, opened 2016-08-19
DataSet and MultiDataSet: add write compressed options


AlexDBlack:
Idea: we have DataSet.read and DataSet.write. In addition to these, it would be good to have the following methods:

DataSet.writeCompressed: Write to the stream using compressed INDArrays.
DataSet.readCompressed: Read the DataSet into memory, keeping compression (if any)
DataSet.read should handle reading compressed DataSet objects (i.e., those written using writeCompressed) but should deserialize into a non-compressed format.

Use cases here:
- Spark, for reduced network transfer (save compressed -> transfer less data over network -> auto decompress)
- For saving space when storing serialized DataSet objects

AlexDBlack:
Still wanted. Low priority, however.

PR #1769, opened 2017-04-13
Enable default compression for DataSet save().

crockpotveggies:
richard-hendricks

The default signature for DataSet.save() now has compression enabled by default. 

Addresses https://github.com/deeplearning4j/nd4j/issues/1224.

## How was this patch tested?

Ran unit tests in addition to tested in a real-world project. Float16 was used for compression since it appears to be the best compression/lossless algorithm.

Weissman score is off the charts!

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/DataSet.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/DataSet.java
@@ -68,7 +68,7 @@ public class DataSet implements org.nd4j.linalg.dataset.api.DataSet {
 
     private List<Serializable> exampleMetaData;
 
-    private String defaultCompression = "GZIP";
+    private String defaultCompression = "FLOAT16";
     private transient boolean preProcessed = false;
 
     public DataSet() {
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/DataSet.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/DataSet.java
@@ -68,6 +68,7 @@ public class DataSet implements org.nd4j.linalg.dataset.api.DataSet {
 
     private List<Serializable> exampleMetaData;
 
+    private String defaultCompression = "GZIP";
     private transient boolean preProcessed = false;
 
     public DataSet() {
@@ -442,10 +443,74 @@ public class DataSet implements org.nd4j.linalg.dataset.api.DataSet {
         }
     }
 
+    public void save(OutputStream to, String compressionAlgorithm) {
+        byte included = 0;
+        if (features != null)
+            included |= BITMASK_FEATURES_PRESENT;
+        if (labels != null) {
+            if (labels == features) {
+                //Same object. Don't serialize the same data twice!
+                included |= BITMASK_LABELS_SAME_AS_FEATURES;
+            } else {
+                included |= BITMASK_LABELS_PRESENT;
+            }
+        }
+        if (featuresMask != null)
+            included |= BITMASK_FEATURE_MASK_PRESENT;
+        if (labelsMask != null)
+            included |= BITMASK_LABELS_MASK_PRESENT;
+
+
+        try {
+            BufferedOutputStream bos = new BufferedOutputStream(to);
+            DataOutputStream dos = new DataOutputStream(bos);
+            dos.writeByte(included);
+
+            INDArray compressed;
+            if (features != null) {
+                compressed = Nd4j.getCompressor().compress(features, compressionAlgorithm);
+                Nd4j.write(compressed, dos);
+            }
+            if (labels != null && labels != features) {
+                compressed = Nd4j.getCompressor().compress(labels, compressionAlgorithm);
+                Nd4j.write(compressed, dos);
+            }
+            if (featuresMask != null) {
+                compressed = Nd4j.getCompressor().compress(featuresMask, compressionAlgorithm);
+                Nd4j.write(compressed, dos);
+            }
+            if (labelsMask != null) {
+                compressed = Nd4j.getCompressor().compress(labelsMask, compressionAlgorithm);
+                Nd4j.write(compressed, dos);
+            }
+
+            dos.flush();
+            dos.close();
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+    }
 
     @Override
     public void save(OutputStream to) {
+        save(to, defaultCompression);
+    }
 
+    public void save(File to, String compressionAlgorithm) {
+        try (FileOutputStream fos = new FileOutputStream(to, false);
+                        BufferedOutputStream bos = new BufferedOutputStream(fos)) {
+            save(bos, compressionAlgorithm);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public void save(File to) {
+        save(to, defaultCompression);
+    }
+
+    public void saveUncompressed(OutputStream to) {
         byte included = 0;
         if (features != null)
             included |= BITMASK_FEATURES_PRESENT;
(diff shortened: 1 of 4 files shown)


crockpotveggies:
First question: do we want/need to update our interface API?

agibsonccc:
@crockpotveggies yeah please do

agibsonccc:
We *could* just dump a warning from save on that datatype?

AlexDBlack:
I'd be fine with adding a one-off warning (once per JVM). "Default DataSet saving configuration is set to FP16. This may result in precision loss. Use .save(...) to specify compression codec, or set default compression codec using Y. Read Z for more details."

crockpotveggies:
Hmmm, I'm getting some strange results when loading my presaved dataset w/ Parallel Wrapper, what's wrong here?

The score is also insanely large. On a single worker, it's 9.x as it should be.

```
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[main] INFO org.deeplearning4j.parallelism.ParallelWrapper - Averaged score: 179947.4941971561
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
[AsyncIterator thread] WARN org.nd4j.linalg.jcublas.buffer.BaseCudaDataBuffer - Loading a data stream with type different from what is set globally. Expect precision loss
```

crockpotveggies:
@raver119 any ideas on the problem above? ^^^^

raver119:
No, poke me in pm & give something to run.

crockpotveggies:
Narrowed this down to incompatibility w/ PW+compression+workspaces. @raver119 will fix before merging this.

raver119:
Making default save() call to save compressed FP16 datasets is terrible idea. It works ok only for images, for anything else, like nlp tasks - it's just disaster. Compression should be optional. I'm reverting default save() signature to work without compression.

raver119:
All we need - to provide option to save compressed datasets. But enforcing lossy compression by default - is terrible UX. I can already see users popping up with questions :)

raver119:
@crockpotveggies once you merge this PR, please merge this one: https://github.com/deeplearning4j/deeplearning4j/pull/3257

It adds decompression hook into ADSI/AMDSI

AlexDBlack:
hm... so thinking about FP16 again: 6.1e-5 is smallest positive value and 2048 is the largest integer that can be represented exactly...
65504 is the largest value that can be represented at all. Given DataSets will often be saved pre-normalization, it's pretty much guaranteed that many users will hit this (and not even realize it). That's a pretty big downside for only a 2x space saving (it might be fine for many use cases - but users should probably have to make that decision for themselves)

crockpotveggies:
Let's decide the "best" fast/lossless compression before merging this and/or abandoning this? The amount of space uncompressed data uses is fairly big. I think in real world situations for many big datasets, it might not be practical to use uncompressed saving. I'm happy to investigate snappy and lz4.

raver119:
The only lossless compression that's implemented atm is gzip, and i doubt it's good thing perf-wise. But if we need something, we could make fast impl in native, i remember Adam included zlib in libnd4j anyway.

crockpotveggies:
Performance on GZIP was *fairly bad* when I tested it. At least 10X than other methods. Yes can we look at a native impl?

crockpotveggies:
What about Zstandard? https://code.facebook.com/posts/1658392934479273/smaller-and-faster-data-compression-with-zstandard/

agibsonccc:
Just of note - I got rid of zlib it was causing too many problems. zlib was only used for saving/loading compressed numpy arrays which we weren't using anyways.

crockpotveggies:
Okay going with lz4. @raver119 you were concerned that lz4 was lossy? But the algo claims it isn't?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/DataSet.java
```diff
@@ -442,10 +443,74 @@ public void load(File from) {
...
+            dos.flush();
+            dos.close();
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+    }
 
     @Override
     public void save(OutputStream to) {
+        saveUncompressed(to);
```
huitseeker: Should this really be the default ??

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/DataSet.java
```diff
@@ -442,10 +443,74 @@ public void load(File from) {
         }
     }
 
+    public void save(OutputStream to, String compressionAlgorithm) {
```
huitseeker: This needs javadoc, not. on the value of `compressionAlgorithm`, please.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/MultiDataSet.java
```diff
@@ -674,6 +674,9 @@ public int hashCode() {
     public long getMemoryFootprint() {
         long reqMem = 0;
 
+        // we need to decompress all arrays before checking its size
```
huitseeker: Grammar.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/DataSet.java
```diff
@@ -324,4 +324,9 @@
      * @return
      */
     long getMemoryFootprint();
+
+    /**
+     * Thia method checks, if dataset is compressed, and applies decompression to it
```
huitseeker: remove commas.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/MultiDataSet.java
```diff
@@ -172,4 +172,9 @@
 
     long getMemoryFootprint();
 
+    /**
+     * Thia method checks, if dataset is compressed, and applies decompression to it
```
huitseeker: Ugh, commas.

agibsonccc:
Closing due to inactivity.

PR #1809, opened 2017-05-03
Example of mandatory checkstyle rules.

priimak:


huitseeker:
We already have a style checker (https://github.com/revelc/formatter-maven-plugin) that checks for google-java-format, with a few exceptions (e.g. we indent to 4 characters). The nice thing is it also comes with IDE support and an automatic formatter (`mvn formatter:format`) that will format according to that style — rather than fail a build. We have tried failing the build: in our fast increments it's proven to be too disruptive a pattern for now.

So, given all of this, it's difficult to keep this PR open : we sincerely appreciate the intent, It's just a feature that is covered for now.

PR #1762, opened 2017-04-10
Fix build for CUDA 7.5 missing cusolverDnSorgqr and cusolverDnDorgqr

saudet:
Fixes #1761 

## What changes were proposed in this pull request?

Calls cusolverDnSorgqr and cusolverDnDorgqr via reflection to fix build with CUDA 7.5.

## How was this patch tested?

Build passes on both CUDA 7.5 and 8.0, but unit tests pass only on CUDA 8.0, as expected.

PR #1834, opened 2017-05-20
Added syev code

rcorbish:
## Issue 927 ( unsupported operation in Eigen values )

Added symmetric eigen -value and -vector code. Main library interfaces are unchanged. Just implemented eigen  for symmetric. CPU & CUDA 8.0 tested

## Test cases included

verified with matlab

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/Lapack.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/blas/Lapack.java
@@ -64,6 +64,20 @@ public interface Lapack {
     public void potrf(INDArray A, boolean lower ) ;
 
 
+    /**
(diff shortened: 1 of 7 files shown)


huitseeker:
Sorry, had to revert this to allow our builds to pass : https://github.com/deeplearning4j/nd4j/pull/1836
Feel free to reopen @rcorbish I'll be glad to merge this once it passes on Cuda 7.5
You can use the `./change-cuda-version.sh` script at the root of the repo to configure the repo for 7.5.

saudet:
@rcorbish Looks like cusolverDnSsyevd and cusolverDnSsyevd isn't available at all on CUDA 7.5, so we'll need to do something like I've done for cusolverDnSorgqr and cusolverDnDorgqr here:
https://github.com/deeplearning4j/nd4j/pull/1762/files

rcorbish:
OK - will take a look at this

rcorbish:
OK - I think I have it. Need to test on 7.5. Should be OK by the weekend

saudet:
Great! Looking forward to the patch :)

PR #1840, opened 2017-05-26
Databuffer reallocation

audreyloeffel:

The databuffers now can be reallocated. 
-> The current pointer and indexer are replaced by new ones.

## How was this patch tested?
New unit tests that test the reallocation within and without a workspace.


saudet:
This looks good for CPU mode, but I think we're going to need to invalidate the CUDA buffers somehow as well? @raver119

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/api/buffer/DoubleDataBufferTest.java
```diff
@@ -212,7 +216,30 @@ public void testOffset() {
...
+    @Test
+    public void testReallocationWorkspace(){
+        WorkspaceConfiguration initialConfig = WorkspaceConfiguration.builder()
+                .initialSize(10 * 1024L * 1024L)
+                .policyAllocation(AllocationPolicy.STRICT)
+                .policyLearning(LearningPolicy.NONE)
+                .build();
+        MemoryWorkspace  workspace = Nd4j.getWorkspaceManager().getAndActivateWorkspace(initialConfig, "SOME_ID");
+
+        DataBuffer buffer = new DoubleBuffer(new double[]{1, 2, 3, 4}, workspace);
```
audreyloeffel: With `DataBuffer buffer = Nd4j.createBuffer(new double[]{1, 2, 3, 4});` ?
It doesn't work : `Assert.assertTrue(buffer.isAttached());` fails with this

How can the buffer be allocated within the workspace ?

audreyloeffel:
True. I could override the method in CudaBaseDataBuffer, no ?

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/api/buffer/DoubleDataBufferTest.java
raver119: Appropriate signature should be added? ND4j code isn't holy cow, if some signature is missing (like workspace use for this given use case) - we usually just add desired functionality :)

raver119: Alternatively - you can poke me, and i'll add it for you. That works too :)

saudet:
@raver119 Would there be a way to free the CUDA memory, so we can have it allocated + data copied from CPU memory, as if it were a new DataBuffer?

raver119:
@saudet sure, it just depends on how they were allocated there's 3 different scenarios:
1) memory was allocated via workspace. So nothing to do here, this chunk is part of workspace, and will be released when workspace gets released.
2) memory was allocated via Ndj methods. In this case i'll need to add some signature, because at this moment such deallocation is protected within AtomicAllocator, due to cache handling for memory chunks.
3) Memory was allocated manually via NativeOps.mallocHost()/mallocDevice() (or Nd4j.getMemoryManager() helper, it's just shortcut after all). In this case memory can be released directly via NativeOps.freeHost()/freeDevice() calls, or corresponding Nd4j.getMemoryManager() helper methods, which are shortcuts as well.

raver119:
Ah, plus there's important use case: device pointer can be null, so can't be released. I.e. when there was device OOM, and host memory use was enforced for a given buffer.

nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/jita/allocator/impl/AtomicAllocator.java
```diff
@@ -364,6 +364,23 @@ public Integer getDeviceId(INDArray array) {
 
 
     /**
+     * This method releases memory allocated for this allocation point
+     * @param point
+     */
+    public void freeMemory(AllocationPoint point) {
```
huitseeker: I'm wondering if it's possible to forget to deallocate on the device because of a wrong `AllocationStatus` here. /newbieComment

nd4j-buffer/src/main/java/org/nd4j/linalg/api/memory/pointers/ImmortalFloatPointer.java
```diff
@@ -29,6 +29,6 @@ public void deallocate() {
     @Override
     public void deallocate(boolean deallocate) {
         log.info("deallocate 2");
-      //  super.deallocate(deallocate);
+        //  super.deallocate(deallocate);
```
huitseeker: I'd either remove commented code or add a line of natural text to indicate why it should stay, here.

nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/jita/allocator/impl/AtomicAllocator.java
raver119: AllocationStatus can't be wrong, that's ground truth. If it's wrong - disaster is a matter of time. Short time though :)

nd4j-buffer/src/main/java/org/nd4j/linalg/api/memory/pointers/ImmortalFloatPointer.java
raver119: It's ok to remove, just debugging remnants .

huitseeker:
@audreyloeffel 👍 :)

PR #1841, opened 2017-05-27
Added syev for lapack - Issue 1834 ( was reverted with bad compile due to cuda 7.5 incompat. )

rcorbish:

Added in reflection to call cuda 8.0 only functions

## How was this patch tested?

Added test case TestEigen
Verified compiling vs cuda 7.5 and 8.0
Checked library failed when methods not present

Please review


saudet:
Looks good to me, @huitseeker ?

rcorbish:
Will we be supporting cuda 9 when it arrives ? I can fax the 7.5 chaps to let them know :)

agibsonccc:
@rcorbish yes could you look at the symbols there to see what we're getting in to here? I'm going to merge this for 7.5 for now. Usually recent nd4j/dl4j releases will only support the 2 most recent versions. We will support 7.5 for this next release.

rcorbish:
I was just being silly. If you say nd4j needs to support version x.x.x it's my responsibility to make sure my check-ins do. It took some time to build a docker image that I can test on :)

PR #1856, opened 2017-06-03
Remove unused maven dependencies + minor goodies

huitseeker:
Partial completion of https://github.com/deeplearning4j/deeplearning4j/issues/3470

- remove unused maven dependencies
- silence Eclipse on lint-maven-plugin
- bump Spark 1.6.2 -> 1.6.3

## How was this patch tested?

- does not break more tests in test suite. (though tests crash atm https://gist.github.com/61fd5c0832164411724d9c5ef9a79506)
- test suite for Datavec, dl4j, dl4j-examples runs ok (or do not [break more](https://github.com/deeplearning4j/deeplearning4j/issues/2909#issuecomment-307997565) in the case of dl4j)

### DIFF
--- a/nd4j-backends/nd4j-backend-impls/nd4j-cuda-platform/pom.xml
+++ b/nd4j-backends/nd4j-backend-impls/nd4j-cuda-platform/pom.xml
@@ -16,41 +16,6 @@
     </properties>
 
     <dependencies>
-        <dependency>
-            <groupId>org.bytedeco.javacpp-presets</groupId>
-            <artifactId>cuda-platform</artifactId>
-            <version>${cuda.version}-${cudnn.version}-${javacpp-presets.version}</version>
-        </dependency>
-
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>${nd4j.backend}</artifactId>
-            <version>${project.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>${nd4j.backend}</artifactId>
-            <version>${project.version}</version>
-            <classifier>${javacpp.platform.linux-x86_64}</classifier>
-        </dependency>
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>${nd4j.backend}</artifactId>
-            <version>${project.version}</version>
-            <classifier>${javacpp.platform.macosx-x86_64}</classifier>
-        </dependency>
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>${nd4j.backend}</artifactId>
-            <version>${project.version}</version>
-            <classifier>${javacpp.platform.windows-x86_64}</classifier>
-        </dependency>
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>${nd4j.backend}</artifactId>
-            <version>${project.version}</version>
-            <classifier>${javacpp.platform.linux-ppc64le}</classifier>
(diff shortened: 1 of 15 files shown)


saudet:
If we remove any of the ones that are related to JavaCPP, a whole lot of things are going to break. Those are not "unused"...

nd4j-backends/nd4j-backend-impls/nd4j-cuda-platform/pom.xml
```diff
@@ -16,41 +16,6 @@
     </properties>
 
     <dependencies>
-        <dependency>
```
AlexDBlack: As already commented: we definitely need the platform/classifier dependencies.

nd4j-backends/nd4j-backend-impls/nd4j-native-platform/pom.xml
```diff
@@ -15,53 +15,6 @@
     </properties>
 
     <dependencies>
-        <dependency>
```
AlexDBlack: As per last comment.

nd4j-common/pom.xml
```diff
@@ -61,12 +50,6 @@
 
         <!-- used to alleviate licensing issues with findbugs -->
```
AlexDBlack: Need to remove comment also.

nd4j-serde/nd4j-camel-routes/nd4j-kafka/pom.xml
```diff
@@ -17,13 +17,6 @@
     </properties>
 
     <dependencies>
-
-        <dependency>
-            <groupId>org.apache.kafka</groupId>
-            <artifactId>kafka_2.11</artifactId>
```
AlexDBlack: I'd definitely double check tests for this one. I'm wondering if camel-kafka has a ```<scope>provided</scope>``` in there, so an explicit dependency is required, at least for tests?
This might also mean that users need to explicitly add a kafka dependency (which isn't necessarily bad, but might be a change in behaviour)

nd4j-backends/nd4j-backend-impls/nd4j-cuda-platform/pom.xml
huitseeker: Yep, those modifications are tool-generated. The fun thing, though, is how far I can go (e.g. running downstream tests) with it.

nd4j-backends/nd4j-backend-impls/nd4j-native-platform/pom.xml
huitseeker: 👍

nd4j-backends/nd4j-backend-impls/pom.xml
```diff
@@ -18,11 +18,6 @@
         <module>nd4j-cuda-platform</module>
     </modules>
     <dependencies>
-        <dependency>
```
huitseeker: This one makes sense (the module has the dependency already).

nd4j-serde/nd4j-camel-routes/nd4j-kafka/pom.xml
huitseeker: Nope, it pulls it https://github.com/apache/camel/blob/master/components/camel-kafka/pom.xml#L53

huitseeker:
@agibsonccc @AlexDBlack, @saudet I have removed the controversial elements, then I haven't found a way to make this one fail for a while.

saudet:
Thanks! Looks good to me now.

Issue #1859, opened 2017-06-06
This project should provide a mixture density network.


jarney:
#### Issue Description

The use of an MSE or L2 cost function means that the network will effectively converge on the mean of the dataset. For some datasets, this is not a desirable behavior and instead, one would like to find the parameters of the distribution of the data.  In particular, it is often desirable to converge on a mixture of gaussian distributions which describe the statistical properties of the data.

For this purpose, nd4j should provide a Mixture Density loss function as described by Bishop[1994].
http://publications.aston.ac.uk/373/1/NCRG_94_004.pdf

#### Version Information

Please indicate relevant versions, including, if relevant:

* Deeplearning4j version: 0.8.1+
* platform information (OS, etc): Any
* CUDA version, if used : None.
* NVIDIA driver version, if in use : None

#### Contributing

I propose to contribute such a loss function after appropriate discussion and review.

PR #1860, opened 2017-06-06
Added a mixture density loss function and corresponding unit-test

jarney:

See issue #1859

The use of an MSE or L2 cost function means that the network will effectively converge on the mean of the dataset. For some datasets, this is not a desirable behavior and instead, one would like to find the parameters of the distribution of the data. In particular, it is often desirable to converge on a mixture of gaussian distributions which describe the statistical properties of the data.

For this purpose, nd4j should provide a Mixture Density loss function as described by Bishop[1994].
http://publications.aston.ac.uk/373/1/NCRG_94_004.pdf

## How was this patch tested?

Unit test was provided and a corresponding example (see PR for dl4j-examples) was used to verify the convergence of the loss function in a controlled setting.  Note that precision has NOT been verified on a CUDA platform because I lack the requisite hardware to test on that platform.  All testing was performed using the native-cpu platform under Ubuntu 16.04.


nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/lossfunctions/TestLossMixtureDensity.java
```diff
@@ -0,0 +1,229 @@
...
+    // identically zero which seems like it's rounding error.
+    // It would be nice to identify the source of the rounding error
+    // so that we can get better fidelity.
+    
+    // This seems to be the precision "sweet spot".
+    // This particular loss function is very sensitive
+    // to floating point arithmetic drift because there are many
+    // operations such as the exponent and the logarithm together
+    // with a bunch of sums, so floating point precision errors is difficult.
+    public static final double epsilon = 1e-2;
```
huitseeker: This is a bit high for comfort — is there more precision to be gained by reusing existing operators as much as possible?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+    
+    /**
+     * This is a 'softmax' where each output yi = e^x divided by
+     * the sum of all 'sum(yi)'.
+     * @param nMixturesPerLabel Number of gaussian mixtures for each label.
+     * @param input Input alpha to softmax on.
+     * @return The softmax of the input (as a rank-3 tensor where the 
+     *         shape is [nSamples, nLabelsPerSample, nMixturesPerLabel].
+     */
+    private static INDArray softmax(int nMixturesPerLabel, INDArray input) {
```
agibsonccc: Yeah agreed with @huitseeker please use our built in softmax.I also wonder about logsoftmax as well? Maybe we should make this activation configurable.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+     * @param labels Labels give the sample output that the network should
+     *               be trying to converge on.
+     * @param preOutput The output of the last layer (before applying the activation function).
+     * @param activationFn The activation function of the current layer.
+     * @param mask Mask to apply to score evaluation (not supported for this cost function).
+     * @return 
+     */
+    @Override
+    public INDArray computeScoreArray(INDArray labels, INDArray preOutput, IActivation activationFn, INDArray mask) {
+        int nSamples = labels.rows();
```
agibsonccc: Are we only assuming 2d here? Or should it be time series? I"d be curious here because we do have the masking here. I'd also consider maybe changing the rows to labels.size(0) if you do.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+        // See http://www.plsyard.com/dealing-overflow-and-underflow-in-softmax-function/
+        // this post for why we calculate the pi_i in this way.
+        // With the exponential function here, we have to be very careful
+        // about overflow/underflow considerations even with
+        // fairly intermediate values.  Subtracting the max
+        // here helps to ensure over/underflow does not happen here.
+        // This isn't exactly a softmax because there's an 'alpha' coefficient
+        // here, but the technique works, nonetheless.
+        INDArray variance = Transforms.pow(mdc.sigma, 2);
+        INDArray minustwovariance = variance.mul(2).negi();
```
agibsonccc: I wonder if we could use dl4j's built in variance here?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+        INDArray normalPart = mdc.alpha.div(Transforms.pow(mdc.sigma.mul(SQRT_TWO_PI), mLabelWidth));
+        INDArray exponent = labelsMinusMuSquared.div(minustwovariance);
+        INDArray exponentMax = exponent.max(1);
+        for (int k = 0; k < mMixtures; k++) {
+            exponent.get(new INDArrayIndex[]{NDArrayIndex.all(), NDArrayIndex.point(k)}).subi(exponentMax);
+        }
+        INDArray pi = Transforms.exp(exponent).mul(normalPart);
```
agibsonccc: Use muli here instead. You already did a dup() with transforms.exp. the i means multiply in place.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+        INDArray piDivisor = pi.sum(1);
+        for (int k = 0; k < mMixtures; k++) {
+            pi.get(new INDArrayIndex[]{NDArrayIndex.all(), NDArrayIndex.point(k)}).divi(piDivisor);
+        }
+
```
agibsonccc: I wonder if there would be a way of using NDArrayIndex.interval rather than a for loop here?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+    
+    /**
+     * This method calculates 'phi' which is the probability
+     * density function (see Bishop 23)
+     * @param diffSquared This is the 'x-mu' term of the Gaussian distribution (distance between 'x' and the mean value of the distribution).
+     * @param sigma This is the standard deviation of the Gaussian distribution.
+     * @return This returns an array of shape [nsamples, nlabels, ndistributions] which contains the probability density (phi) for each of the
+     *         samples * labels * distributions for the given x, sigma, mu.
+     */
+    INDArray phi(INDArray diffSquared, INDArray sigma) {
```
agibsonccc: private or protected here?

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/lossfunctions/TestLossMixtureDensity.java
```diff
@@ -0,0 +1,229 @@
...
+            { { 0.50, 0.96, 0.31}, {0.18, 0.85, 0.76, 0.80, 0.60, 0.93, 0.38, 0.65, 0.27, 0.02} },
+            { { 0.51, 0.85, 0.51}, {0.77, 0.65, 0.39, 0.94, 0.83, 0.70, 0.63, 0.83, 0.81, 0.43} },
+            { { 0.76, 0.60, 0.59}, {0.62, 0.89, 0.11, 0.14, 0.04, 0.70, 0.05, 0.07, 0.23, 0.54} },
+            { { 0.20, 0.81, 0.39}, {0.12, 0.19, 0.57, 0.00, 0.00, 0.87, 0.41, 0.08, 0.62, 0.38} },
+            { { 0.08, 0.17, 0.48}, {0.79, 0.78, 0.81, 0.02, 0.56, 0.84, 0.45, 0.24, 0.84, 0.16} },
+            { { 0.55, 0.94, 0.55}, {0.43, 0.49, 0.96, 0.46, 0.64, 0.35, 0.75, 0.18, 0.47, 0.59} },
+        };
+
+        int totalNFailures = 0;
+        for (int i = 0; i < labelPreOutputTestData.length; i++) {
```
agibsonccc: Maybe you could use our GradientCheckUtil here similar to https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/GradientCheckTests.java

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.factory.Nd4j;
+import org.nd4j.linalg.indexing.INDArrayIndex;
+import org.nd4j.linalg.indexing.NDArrayIndex;
+import org.nd4j.linalg.lossfunctions.ILossFunction;
+import org.nd4j.linalg.ops.transforms.Transforms;
+
+/**
+ * This is a cost function associated with a mixture-density network.
+ * For background, this is inspired by Bishop's work pioneering the mixture
```
jarney: Will do, thanks.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+    
+    /**
+     * This class is a data holder for the mixture density
+     * components for convenient manipulation.
+     * These are organized as rank-3 matrices with shape
+     * [nSamples, nLabelsPerSample, nMixturesPerLabel]
+     * and refer to the 'alpha' (weight of that gaussian), 'mu' (mean for that
+     * gaussian), and 'sigma' (standard-deviation for that gaussian).
+     */
+    public static class MixtureDensityComponents {
```
jarney: Will do, thanks.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
jarney: Will change this to use existing SoftMax.
As to the question on configurability, in order for the derivations of Bishop to be valid, the alpha must sum to identically 1.  Making this configurable would invalidate some of the other derivatives and it's not clear what the result would be, hence I would prefer to keep this specifically as SoftMax.

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/lossfunctions/TestLossMixtureDensity.java
jarney: I agree that this is high and I would like precision to be better.  I set this to 1e-2 to ensure pass on the unit-tests.

In my build, many of the other loss functions show similar behavior, but for some reason, don't fail the unit-tests.  Note lack of Assert() here:
https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/lossfunctions/LossFunctionGradientChecks.java#L159

In my built, this results in most of the loss functions giving FAIL for at least some of the cases:

I'm not really sure what the cause is and I'm not sure if there's already an issue logged against this or if this is something people are already aware of.
Starting test: LossMSE(), identity, input shape = [1, 3]
Param 4 FAILED: grad= 0.12049015427391895, numericalGrad= 0.09125180355784579, relErrorPerc= 32.04139488326777, scorePlus=1.2416819773168561, scoreMinus= 1.241681794813249
Param 4 passed: grad= -0.14576807108993803, numericalGrad= -0.1480154177002646, relError= 1.518319270548909, scorePlus=1.2416817380495904, scoreMinus= 1.2416820340804258
Param 4 FAILED: grad= 0.02527791681601911, numericalGrad= 0.05676361425344112, relErrorPerc= 55.46809844919853, scorePlus=1.2416819428286618, scoreMinus= 1.2416818293014333
Gradient check failed for loss function LossMSE(); total num failures = 2
DONE
Starting test: LossMSE(), identity, input shape = [3, 3]
Param 5 FAILED: grad= -0.5873957915863871, numericalGrad= -1.8712516975938343, relErrorPerc= 68.60947181281412, scorePlus=7052.590421763204, scoreMinus= 7052.590425505708
Param 5 FAILED: grad= 3.981470399920717, numericalGrad= 4.710964276455343, relErrorPerc= 15.485022465156899, scorePlus=7052.590428345421, scoreMinus= 7052.5904189234925
Param 5 FAILED: grad= -3.3940746083343303, numericalGrad= -2.8397130336088594, relErrorPerc= 19.52174632311204, scorePlus=7052.590420794742, scoreMinus= 7052.590426474168
Param 5 FAILED: grad= -0.1458934913657659, numericalGrad= -0.38700090954080224, relErrorPerc= 62.301512019990724, scorePlus=7052.590423247456, scoreMinus= 7052.590424021458
Param 5 FAILED: grad= -0.0065448732525173045, numericalGrad= 0.08938513929024339, relErrorPerc= 107.32210443982795, scorePlus=7052.590423723842, scoreMinus= 7052.590423545072
Param 5 FAILED: grad= 0.15243836461828322, numericalGrad= 0.29761577025055885, relErrorPerc= 48.78014545736358, scorePlus=7052.590423932073, scoreMinus= 7052.590423336841
Param 5 FAILED: grad= -5.490417982591264, numericalGrad= -8.184971648006467, relErrorPerc= 32.92074525476809, scorePlus=7052.590415449482, scoreMinus= 7052.590431819425
Param 5 FAILED: grad= 2.1234013945022925, numericalGrad= 3.3275196074100677, relErrorPerc= 36.186660184550654, scorePlus=7052.590426961975, scoreMinus= 7052.590420306936
Param 5 FAILED: grad= 3.367016588088971, numericalGrad= 4.857451585849049, relErrorPerc= 30.683476127730874, scorePlus=7052.590428491909, scoreMinus= 7052.590418777006
Gradient check failed for loss function LossMSE(); total num failures = 9
DONE
Starting test: LossL1(), sigmoid, input shape = [1, 3]
Param 6 passed: grad= -0.24688434738311787, numericalGrad= -0.24622887506708935, relError= 0.2662044879382778, scorePlus=2.6500985168042113, scoreMinus= 2.6500990092619614
Param 6 FAILED: grad= -0.22703591494179454, numericalGrad= -0.24680589394421304, relErrorPerc= 8.010335039602914, scorePlus=2.6500985162271937, scoreMinus= 2.6500990098389816
Param 6 FAILED: grad= -0.12886947021740552, numericalGrad= -0.24896490735670795, relErrorPerc= 48.237897627569666, scorePlus=2.650098514068186, scoreMinus= 2.650099011998001
Gradient check failed for loss function LossL1(); total num failures = 2
DONE

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
jarney: In this context, I'm just trying to calculate -2*sigma^2, so I'm not sure about a built-in variance function.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
jarney: Will do.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
jarney: Making this private.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
jarney: I agree, there is probably a more elegant way to do this, but I wasn't exactly sure how.  I'll make another pass at this part and see what I can figure out.

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/lossfunctions/TestLossMixtureDensity.java
jarney: I modeled my changes after https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/lossfunctions/LossFunctionGradientChecks.java#L159 instead.  I would have used exactly this function to test mine except for the already existing failures I observed (see previous comment).  I didn't use the GradientCheckUtil because that's a part of deeplearning4j and the loss functions are in nd4j, so they're not available at that level unless I move the test into another project.

jarney:
Thank you for your careful and thoughtful reviews.  I'll rework accordingly and attempt to resolve the comments.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
jarney: Will change rows() -> size(0) and add masking.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
AlexDBlack: Yeah, built-in softmax is better. Non-configurable makes sense to me for now (I'm not sure what else we'd use that has the "sums to 1" property).

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
AlexDBlack: Assuming 2d is safe here - the other loss functions assume this also.
In terms of time series, we handle that (by reshaping from 3d to 2d) at the DL4J level. It keeps the loss functions themselves simpler.

Masking is definitely necessary though - per-time step masking (i.e., column vector mask array, values 0 if absent, 1 if present) is definitely necessary. 
https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossL1.java#L66-L68
That is used to allow variable length time series, many-to-one etc with RNNs (but again, it's only 2d data from the perspective of the loss function).

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+        // This computes pi_i, see Bishop equation (30).
+        // See http://www.plsyard.com/dealing-overflow-and-underflow-in-softmax-function/
+        // this post for why we calculate the pi_i in this way.
+        // With the exponential function here, we have to be very careful
+        // about overflow/underflow considerations even with
+        // fairly intermediate values.  Subtracting the max
+        // here helps to ensure over/underflow does not happen here.
+        // This isn't exactly a softmax because there's an 'alpha' coefficient
+        // here, but the technique works, nonetheless.
+        INDArray variance = Transforms.pow(mdc.sigma, 2);
```
AlexDBlack: Do mdc.sigma.mul(mdc.sigma) - mathematically the same, but it's likely faster than a pow op in practice.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+        INDArray minustwovariance = variance.mul(2).negi();
+        INDArray normalPart = mdc.alpha.div(Transforms.pow(mdc.sigma.mul(SQRT_TWO_PI), mLabelWidth));
```
AlexDBlack: This is getting really picky with the performance, but ```Transforms.pow(mdc.sigma.mul(SQRT_TWO_PI), mLabelWidth).rdivi(mdc.alpha);``` is equivalent but faster (one less temp/working array)

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
AlexDBlack: If piDivisor is a column vector (it should be after a sum(1)) and you want to apply element-wise division to each column, you can use ```pi.diviColumnVector(piDivisor)```

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+        INDArray piDivisor = pi.sum(1);
+        for (int k = 0; k < mMixtures; k++) {
+            pi.get(new INDArrayIndex[]{NDArrayIndex.all(), NDArrayIndex.point(k)}).divi(piDivisor);
+        }
+
+        // See Bishop equation (35)
+        //INDArray dLdZAlpha = Nd4j.zeros(nSamples, nLabelsPerSample, mMixturesPerLabel); //mdc.alpha.sub(pi);
+        INDArray dLdZAlpha = mdc.alpha.sub(pi);
+        // See Bishop equation (38)
+        INDArray dLdZSigma = (labelsMinusMuSquared.div(variance).sub(mLabelWidth)).mul(-1).mul(pi);
```
AlexDBlack: For performance, we can reuse arrays:
``` (labelsMinusMuSquared.div(variance).sub(mLabelWidth)).mul(-1).mul(pi);``` becomes
``` (labelsMinusMuSquared.div(variance).subi(mLabelWidth)).muli(-1).muli(pi);``` (only the first div op is a copy op to give us a new result array; the rest do in-place modifications on that)

File: nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java [hunk 14]
```diff
@@ -0,0 +1,399 @@
...
+        // See Bishop equation (39)
+        INDArray dLdZMu = labelsMinusMu.neg();
+        for (int k = 0; k < mLabelWidth; k++) {
+            dLdZMu.get(new INDArrayIndex[]{NDArrayIndex.all(), NDArrayIndex.all(), NDArrayIndex.point(k)}).divi(variance).muli(pi);
```
AlexDBlack: Not a big deal for now, but we could look at broadcast ops for this if we want to optimize it a bit more.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+     * This method calculates 'phi' which is the probability
+     * density function (see Bishop 23)
+     * @param diffSquared This is the 'x-mu' term of the Gaussian distribution (distance between 'x' and the mean value of the distribution).
+     * @param sigma This is the standard deviation of the Gaussian distribution.
+     * @return This returns an array of shape [nsamples, nlabels, ndistributions] which contains the probability density (phi) for each of the
+     *         samples * labels * distributions for the given x, sigma, mu.
+     */
+    INDArray phi(INDArray diffSquared, INDArray sigma) {
+        // 1/sqrt(2PIs^2) * e^((in-u)^2/2*s^2)
+        INDArray minustwovariance = Transforms.pow(sigma, 2).mul(2).negi();
```
AlexDBlack: use ```sigma.mul(sigma).muli(-2)```

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+
+        INDArray normalPart = Transforms.pow(sigma.mul(SQRT_TWO_PI), (double)mLabelWidth);
```
AlexDBlack: ```Transforms.pow(sigma.mul(SQRT_TWO_PI), (double)mLabelWidth, false);``` to avoid an extra array allocation.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+
+        INDArray exponentPart = Transforms.exp(
+            diffSquared.div(minustwovariance)
+        );
```
AlexDBlack: ```Transforms.exp(diffSquared.div(minustwovariance), false);``` to avoid extra array allocation.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
```diff
@@ -0,0 +1,399 @@
...
+
+        // This is phi_i(x,mu,sigma)
+        INDArray likelihoods = 
+                exponentPart.div(normalPart);
```
AlexDBlack: Use ```divi``` - the exponentPart array isn't used again later, so it's safe to re-use.

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/lossfunctions/TestLossMixtureDensity.java
AlexDBlack: It could just be a single/double precision issue; gradient checks should always be done with double precision.
FWIW our main loss function gradient checks are here: https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-core/src/test/java/org/deeplearning4j/gradientcheck/LossFunctionGradientCheck.java

I'd strongly suggest adding a case for the new loss function to that DL4J unit test (or, something equivalent there) - I'd be uncomfortable with accepting this just based on the 1e-2 level. The rest pass the DL4J loss function gradient checks at the 1e-5 level (and most of them would pass with a much tighter threshold than that).

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/lossfunctions/TestLossMixtureDensity.java
```diff
@@ -0,0 +1,229 @@
...
+    @Test
+    public void testCostAgainstScore() {
+        // Mixture of 2 gaussians with 3 inputs each.
+        ILossFunction lf = LossMixtureDensity.builder()
+                .gaussians(2)
+                .labelWidth(3)
+                .build();
+        IActivation activation = Activation.IDENTITY.getActivationFunction();
+        
+        // This is a bunch of random data
```
AlexDBlack: If this is just uniform random data, Nd4j.rand(...) might be cleaner? We can use Nd4j.getRandom().setSeed(someNumber) for repeatability.

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/lossfunctions/TestLossMixtureDensity.java
jarney: I see, I saw the gradient check in nd4j and assumed that was "the one".  Hopefully this accounts for the precision difference.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMixtureDensity.java
jarney: Thank you, this is exactly the idiom I was searching for.

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/lossfunctions/TestLossMixtureDensity.java
jarney: YES!  Using the other gradient check util with Double gave a MUCH cleaner result...
.d.g.LossFunctionGradientCheck -  ***** Starting test: org.nd4j.linalg.lossfunctions.impl.LossMixtureDensity@815b41f - identity - minibatchSize = 1 *****
o.d.g.GradientCheckUtil - Param 0 (0_W) passed: grad= -0.6053853982355434, numericalGrad= -0.6053853986998092, relError= 3.834465194925549E-10
o.d.g.GradientCheckUtil - Param 1 (0_W) passed: grad= -0.47783684344684657, numericalGrad= -0.4778368429825264, relError= 4.858564173090874E-10

This seems to have accounted for the problem.  Give me a bit of time to take in all the comments and prepare another patch.

jarney:
@AlexDBlack , thank you so much for your review.  I had not realized that I was using single-precision float for my tests.  When I switched to doubles using the unit-test you suggested, the precision problems simply disappeared.  Give me a few days to rework based on the collective review of you and your peers.  Thank you so much for your time and patience with me.

jarney:
For improved gradient checks, please refer to LossFunctionGradientCheck : https://github.com/deeplearning4j/deeplearning4j/pull/3505

jarney:
@AlexDBlack : I have made a pass at the gradient check accuracy as well as many of the optimizations you pointed out.

I think the sticking point I have now is related to the "mask".  Conceptually, I'm having trouble understanding exactly what is being masked here.  Is it the:
* Score shape = (nSamples, 1)
* Labels shape = (nSamples, labelWidth)
* Network Output = (nSamples, (labelWidth+2)*mixtures)

If you could help me understand that, I should be able to close the gap on masking.

AlexDBlack:
@jarney Apologies for the delay here (for future reference, feel free to ping me on gitter - https://gitter.im/AlexDBlack)

I scanned through the PR again, and the unit test added here - https://github.com/deeplearning4j/deeplearning4j/pull/3505/files
Looks good to me. I can see some scope for additional optimizations, but that's not urgent; maybe I'll do a quick pass post-merge.


As for masking: your implementation there looks fine.
To answer your question on masks: it's basically "example is/isn't present for this row". Most common use case: RNNs, with a label at the final step, not *every* step.
What the mask does: removes the contribution of the score for that row - and hence the gradients also - if mask set to 0 for that step.


Next steps: let's get the test PR updated to latest master, and then we can merge this.
https://github.com/deeplearning4j/deeplearning4j/pull/3505

jarney:
@AlexDBlack Thanks.

I've re-based the unit-test, resolving the merge conflict, so I think we're good there.

For the masking issue, I'm still not convinced it's correct as it stands, but you're probably a better judge of that.  The reason I have some concern is that it appears that the masking logic makes the assumption that size(mask) = size(labels) = size(output) which isn't the case here.  It "looks" fine, but I'm not sure it's actually correct in this situation.

As a side note, have you considered moving the masking out of the loss functions?  It would almost seem that the mask isn't really a concern of the loss function, but is instead a concern of the network topology and the loss functions should be relatively agnostic to any masking of the data given to them.

jarney:
I think I've wrapped my head around the masking.  It was not correct as it stood, however, I changed the mask application at the "computeScore" method to mask based on output shape and it now passes the Masking gradient check as well.  See deeplearning4j/deeplearning4j#3505 for the amendment that includes a "Mask" unit-test as well.

Part of the confusion is that in that unit-test, the mask is named "labelMask", however, many places that use the mask, do so based on the output shape, not the label shape.  I think this was my confusion in trying to apply the mask to the label.

AlexDBlack:
I have to disagree with the change you made in the most recent commit.
Previously, you mased the score array - now you mask the output - which is incorrect.

To recap, there's 2 types of masking:
- Per-*example* masking -> mask is a column vector. Mixture density loss function has to support this
- Per-*output* masking -> mask has same shape as output. This doesn't make sense for mixture density, and we shouldn't technically support it here

What the per-example masking should do:
- Set the *score* to 0 for any examples that are masked out - they won't contribute to the overall minibatch score
- Set the *gradients* (dL/dPreActivationFn) to 0 for any examples that are masked out - this is obvious from the score... if activations don't contribute to the score (i.e., example/row is mask) then by definition that derivative must be 0 for those examples/rows


Unless I'm missing something with the math, masking the output (i.e., setting it to 0) does not result in the score (negative log likelihood) being 0 for the masked-out examples, which is the whole point of masking. The *gradient* masking (LossUtil.applyMask(gradients, mask);) looks correct to me, however.

As to why it would pass gradient checks with the output being masked - and not the scores array - is to me a mystery...

jarney:
Thanks for the explanation, it's more clear now.

Ok, I've reverted the masking code to where it was yesterday (masking score instead of output).
I've also withdrawn the unit-test for GradientCheckMasking.testPerOutputMaskingMLP since we only want to support per-example masks.

AlexDBlack:
@jarney Both PRs now merged. Thanks again!

Issue #1878, opened 2017-06-15
Add SELU Activation Function


mdavis95:
https://arxiv.org/pdf/1706.02515.pdf

![screenshot from 2017-06-15 12-21-27](https://user-images.githubusercontent.com/7322079/27191314-6ae78802-51c5-11e7-8778-b8e3dab7c908.png)

α ≈ 1.6732632423543772848170429916717
λ ≈ 1.0507009873554804934193349852946

raver119:
Implemented and merged.

mdavis95:
Thanks!

PR #1879, opened 2017-06-15
SELU

raver119:
SELU, SELUDerivative + ActivationSELU wrappers on java side

As requested here: https://github.com/deeplearning4j/nd4j/issues/1878


maxpumperla:
@raver119 nice commit message :)

Not sure if I understand this correctly, but for now this seems to just be `tan` and derivative `1/cos^2`, not the scaled exponential with the two weird constants. Still WIP?

raver119:
Nah, it's just wrapper for libnd4j part, activation is implemented in native code, and tan/derivative is just copy/paste issue, java op isn't used... Should fix that though, thanks for highlighting

raver119:
https://github.com/deeplearning4j/libnd4j/pull/401

maxpumperla:
interesting, thanks for clarifying this. otherwise lgtm!

PR #1881, opened 2017-06-18
Undersampling with masking for tbptt

eraly:
~Do not merge.~
~Needs javadoc cleanup and tests can be cleanedup~


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/dataset/api/preprocessor/classimbalance/BaseUnderSamplingPreProcessor.java
```diff
@@ -0,0 +1,103 @@
...
+    */
+    private INDArray calculateBernoulli(INDArray minorityLabels, INDArray labelMask, double targetMinorityDist) {
+
+        INDArray minorityClass = minorityLabels.dup().muli(labelMask);
+        INDArray majorityClass = Transforms.not(minorityLabels).muli(labelMask);
+
+        //all minorityLabel class, keep masks as is
+        if (majorityClass.sumNumber().intValue() == 0) return labelMask;
+        //all majority class and set to not mask all majority windows sample majority class by 1-targetMinorityDist
+        if (minorityClass.sumNumber().intValue() == 0 && !maskAllMajorityWindows) return labelMask.muli(1-targetMinorityDist);
```
AlexDBlack: Maybe clarify in the code what's going on here? Initially it looks wrong (as label masks should have values 0 or 1 only). However, we're actually using this as the bernoulli sampling probability distribution, which does make sense.

eraly: Fixed

eraly:
Fixed check for NaNs.

PR #1888, opened 2017-06-22
Minor changes for Spark PS

raver119:
This PR adds changes for new distributed spark impl.

- VoidParameterServer changes
- memset exposed

### DIFF
--- a/nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
+++ b/nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
@@ -7,6 +7,7 @@ import org.apache.commons.math3.util.Pair;
 import org.nd4j.linalg.api.ndarray.INDArray;
 import org.nd4j.linalg.exception.ND4JIllegalStateException;
 import org.nd4j.parameterserver.distributed.conf.VoidConfiguration;
+import org.nd4j.parameterserver.distributed.enums.ExecutionMode;
 import org.nd4j.parameterserver.distributed.enums.NodeRole;
 import org.nd4j.parameterserver.distributed.logic.*;
 import org.nd4j.parameterserver.distributed.logic.completion.Clipboard;
@@ -191,8 +192,7 @@ public class VoidParameterServer {
                 this.transport = transport;
 
                 // first we need to check, if our current IP matches designated shards or backup
-                if (nodeRole == NodeRole.NONE && (voidConfiguration.getForcedRole() == null
-                                || voidConfiguration.getForcedRole() == NodeRole.NONE)) {
+                if (nodeRole == NodeRole.NONE && (voidConfiguration.getForcedRole() == null || voidConfiguration.getForcedRole() == NodeRole.NONE)) {
                     Pair<NodeRole, String> pair = null;
                     if (voidConfiguration.getShardAddresses().size() == 1
                                     && voidConfiguration.getShardAddresses().get(0).contains("127.0.0.1")) {
@@ -239,8 +239,10 @@ public class VoidParameterServer {
                     if (nodeRole == NodeRole.NONE)
                         nodeRole = voidConfiguration.getForcedRole();
 
-                    this.transport.init(voidConfiguration, clipboard, nodeRole, "127.0.0.1",
-                                    voidConfiguration.getUnicastPort(), shardIndex);
+                    // if we're using forced roles here, we'll assume that controllerAddress belongs to this box
+                    String localIp = voidConfiguration.getExecutionMode() == ExecutionMode.MANAGED ? voidConfiguration.getControllerAddress() : "127.0.0.1";
+
+                    this.transport.init(voidConfiguration, clipboard, nodeRole, localIp, voidConfiguration.getUnicastPort(), shardIndex);
                 }
 
 
--- a/nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
+++ b/nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
@@ -476,4 +476,31 @@ public class VoidParameterServer {
 
         return response.getPayload();
     }
+
+    /**
+     * This method sends given message to all Shards
+     *
+     * @param message
+     */
+    public synchronized void sendMessageToAllShards(@NonNull VoidMessage message) {
+        transport.sendMessageToAllShards(message);
+    }
+
+    /**
+     * This method sends given message to all Clients
+     *
+     * @param message
+     */
+    public void sendMessageToAllClients(@NonNull VoidMessage message) {
+        this.sendMessageToAllClients(message, null);
+    }
+
+    /**
+     * This method sends given message to all Clients, excluding
+     *
+     * @param message
+     */
+    public synchronized void sendMessageToAllClients(@NonNull VoidMessage message, Long... exclusions) {
+        transport.sendMessageToAllClients(message);
+    }
 }
--- a/nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
+++ b/nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
@@ -155,6 +155,14 @@ public class VoidParameterServer {
     }
 
     /**
+     * This method returns True if initialization was started AND was finished, false otherwise
+     * @return
+     */
+    public boolean isInit() {
+        return initFinished.get();
+    }
+
+    /**
      * This method starts ParameterServer instance
      *
      * PLEASE NOTE: This method is blocking for first caller only
--- a/nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
+++ b/nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
@@ -444,6 +444,10 @@ public class VoidParameterServer {
         //transport.sendMessage(message);
     }
 
+    public synchronized void execDistributedImmediately(@NonNull TrainingMessage message) {
+        transport.sendMessageToAllShards(message);
+    }
+
     public void execDistributed(@NonNull Frame<? extends TrainingMessage> messages) {
         transport.sendMessage(messages);
     }
--- a/nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
+++ b/nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
@@ -6,6 +6,7 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.math3.util.Pair;
 import org.nd4j.linalg.api.ndarray.INDArray;
 import org.nd4j.linalg.exception.ND4JIllegalStateException;
+import org.nd4j.linalg.factory.Nd4j;
 import org.nd4j.parameterserver.distributed.conf.VoidConfiguration;
 import org.nd4j.parameterserver.distributed.enums.ExecutionMode;
 import org.nd4j.parameterserver.distributed.enums.NodeRole;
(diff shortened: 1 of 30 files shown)


nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/conf/VoidConfiguration.java
```diff
@@ -74,7 +79,8 @@ public void setBackupAddresses(List<String> addresses) {
     }
 
     public void setBackupAddresses(String... Ips) {
-        backupAddresses = new ArrayList<>();
```
agibsonccc: Can we lowercase this?

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/conf/VoidConfiguration.java
```diff
@@ -21,32 +21,36 @@
...
-    private long responseTimeframe;
-    private long responseTimeout;
+    @Builder.Default private long retransmitTimeout = 1000;
+    @Builder.Default private long responseTimeframe = 500;
+    @Builder.Default private long responseTimeout = 30000;
+
+    private String controllerAddress;
 
     public void setStreamId(int streamId) {
         if (streamId < 1)
```
agibsonccc: May want to link to aeron here for those who don't know what  a "stream" is.

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/conf/VoidConfiguration.java
```diff
@@ -61,7 +65,8 @@ public void setShardAddresses(List<String> addresses) {
     }
 
     public void setShardAddresses(String... Ips) {
-        shardAddresses = new ArrayList<>();
+        if (shardAddresses == null)
```
agibsonccc: We should be clear on the format here. Are hostnames also possible?

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/messages/intercom/DistributedInitializationMessage.java
```diff
@@ -60,7 +60,7 @@ public void processMessage() {
             if (voidConfiguration.getExecutionMode() == ExecutionMode.AVERAGING) {
                 // each shard has full own copy
                 columnsPerShard = vectorLength;
-            } else if (voidConfiguration.getExecutionMode() == ExecutionMode.DISTRIBUTED) {
+            } else if (voidConfiguration.getExecutionMode() == ExecutionMode.SHARDED) {
```
agibsonccc: Won't certain configurations be sharded by default? I know I had sharding built in to the media driver communications for say: certain array sizes.

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
```diff
@@ -231,8 +239,10 @@ public void init(@NonNull VoidConfiguration voidConfiguration, @NonNull Transpor
                     if (nodeRole == NodeRole.NONE)
                         nodeRole = voidConfiguration.getForcedRole();
 
-                    this.transport.init(voidConfiguration, clipboard, nodeRole, "127.0.0.1",
-                                    voidConfiguration.getUnicastPort(), shardIndex);
+                    // if we're using forced roles here, we'll assume that controllerAddress belongs to this box
+                    String localIp = voidConfiguration.getExecutionMode() == ExecutionMode.MANAGED ? voidConfiguration.getControllerAddress() : "127.0.0.1";
+
```
agibsonccc: Wouldn't 0.0.0.0 be better here?

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
```diff
@@ -444,6 +454,10 @@ public synchronized void execDistributed(@NonNull TrainingMessage message) {
         //transport.sendMessage(message);
     }
 
+    public synchronized void execDistributedImmediately(@NonNull TrainingMessage message) {
+        transport.sendMessageToAllShards(message);
```
agibsonccc: Would a future be better than synchronized here?

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
```diff
@@ -472,4 +486,31 @@ public INDArray getVector(@NonNull Integer key, int rowIdx) {
...
         return response.getPayload();
     }
+
+    /**
+     * This method sends given message to all Shards
+     *
+     * @param message
+     */
+    public synchronized void sendMessageToAllShards(@NonNull VoidMessage message) {
+        transport.sendMessageToAllShards(message);
```
agibsonccc: Again. Future?

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/conf/VoidConfiguration.java
```diff
@@ -21,32 +22,37 @@
...
+    @Builder.Default private int unicastPort = 49876;
+    @Builder.Default private int multicastPort = 59876;
+    @Builder.Default private int numberOfShards = 1;;
+
+    @Builder.Default private FaultToleranceStrategy faultToleranceStrategy = FaultToleranceStrategy.NONE;
+    @Builder.Default private ExecutionMode executionMode = ExecutionMode.SHARDED;
+
+    @Builder.Default private List<String> shardAddresses = new ArrayList<>();
+    @Builder.Default private List<String> backupAddresses = new ArrayList<>();
+    @Builder.Default private TransportType transportType = TransportType.ROUTED;
```
agibsonccc: May want some javadoc on what "routed" is.

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/transport/RoutedTransport.java
```diff
@@ -160,9 +159,90 @@ public void init(@NonNull VoidConfiguration voidConfiguration, @NonNull Clipboar
...
+                    case NOT_CONNECTED: {
+                        if (!rc.getActivated().get()) {
+                            retr++;
 
+                            if (retr > 20)
+                                throw new ND4JIllegalStateException(
+                                        "Can't connect to Shard: [" + rc.getPublication().channel() + "]");
 
+                            try {
+                                Thread.sleep(voidConfiguration.getRetransmitTimeout());
```
agibsonccc: Maybe we should use aeron's wait mechanisms instead of Thread.sleep?

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/transport/RoutedTransport.java
```diff
@@ -160,9 +159,90 @@ public void init(@NonNull VoidConfiguration voidConfiguration, @NonNull Clipboar
...
+                        } else {
+                            throw new ND4JIllegalStateException("Shards reassignment is to be implemented yet");
+                        }
+                    }
+                    break;
+                    case ADMIN_ACTION:
+                    case BACKPRESSURE: {
+                        try {
+                            Thread.sleep(voidConfiguration.getRetransmitTimeout());
+                        } catch (Exception e) {
```
agibsonccc: Again. Aeron's wait mechanisms rather than Thread.sleep?

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/transport/RoutedTransport.java
```diff
@@ -308,6 +392,16 @@ protected void sendFeedbackToClient(VoidMessage message) {
     }
 
     @Override
+    public int numberOfKnownClients() {
+        return clients.size();
```
agibsonccc: This worries me a bit..I wonder if we should make this a "provider" interface of some kind. I would like the ability to put this in a zookeeper cluster. This would also integrate with SKIL better.

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/transport/RoutedTransport.java
```diff
@@ -489,6 +585,21 @@ public void launch(@NonNull ThreadingModel threading) {
...
 
+
+    @Override
+    public synchronized void addShard(String ip, int port) {
+        Long hash = HashUtil.getLongHash(ip + ":" + port);
+
+        RemoteConnection connection = RemoteConnection.builder().ip(ip).port(port)
+                .publication(aeron.addPublication("aeron:udp?endpoint=" + ip + ":" + port,
+                        voidConfiguration.getStreamId()))
+                .longHash(hash)
```
agibsonccc: Some fault tolerant agent management with some built in fault tolerance would be good.

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/conf/VoidConfiguration.java
raver119: There's no real need for anyone to bother with that tbh. I'll add that, sure, but under no circumstances users will be required to change default values here. Unless they will have stuff in the same JVM process, sure etc.

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
raver119: Sure if that's ok for Aeron, but that won't change anything. In this particular case it literally means "spark local", so all communications will be routed to java Queues, just to avoid redundant serialization which isn't cheap :/

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
raver119: Hm. Good point. It shouldn't be synchronized at all, we're using locks on publisher level.

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/VoidParameterServer.java
raver119: Same answer as above. Draft-time keyword. Will be removed.

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/transport/RoutedTransport.java
raver119: This thing knows nothing about anything beyond it's own state - there's Master node, it always knows about clients, by design - each Client introduces itself first, sends something next. So this is just internal method, used internally. I.e. for messages retransmission.  So it's not management thing or something, it's solely for messages routing.

I.e. for MulticastTransport - this is useless, and isn't used, because messages are sent to everyone at once, and there's no need in such handling.

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/transport/RoutedTransport.java
raver119: Dead Spark Master == dead cluster. And death of any individual node isn't a game changer, all messages are in one of 2 states: delivered to master or lost together with node. In both cases that never changes (or should NOT change) state of any other node in cluster, or master's state.

raver119:
Damn, notifications for review comments should be improved somehow. Seriously. So easy to miss :(

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/transport/RoutedTransport.java
raver119: However, nothing stops us from keeping second "master model" instance somewhere else... Basically any Client has such instance at any moment of time

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/conf/VoidConfiguration.java
raver119: https://github.com/deeplearning4j/nd4j/blob/fb5c71b98ef6fbfbf6df875b0fc6756be7fd7402/nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/enums/TransportType.java

Javadoc for this part is already here

nd4j-parameter-server-parent/nd4j-parameter-server-node/src/main/java/org/nd4j/parameterserver/distributed/transport/RoutedTransport.java
raver119: Aeron's wait mechanics is basically the same 3 things we use:
1) Thread.sleep()
2) LockSupport.parkNanos()
3) busy-lock.

Bit in this particular case we really need sleep, we don't need short retransmission frame in nanos like when. We really just need to sleep some time, to give recipient some free time to solve its issues. But sure, i can use LockSupport here if you want.

PR #1891, opened 2017-06-27
Added Nadam updater.

deil87:

Introduced Nadam updater

## How was this patch tested?

Similarly to other updaters in nd4j and deeplearning4j
Please review

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/NadamUpdater.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/NadamUpdater.java
@@ -1,0 +1,101 @@
+/*-
+ *
+ *  * Copyright 2017 Skymind,Inc.
+ *  *
+ *  *    Licensed under the Apache License, Version 2.0 (the "License");
+ *  *    you may not use this file except in compliance with the License.
+ *  *    You may obtain a copy of the License at
+ *  *
+ *  *        http://www.apache.org/licenses/LICENSE-2.0
+ *  *
+ *  *    Unless required by applicable law or agreed to in writing, software
+ *  *    distributed under the License is distributed on an "AS IS" BASIS,
+ *  *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  *    See the License for the specific language governing permissions and
+ *  *    limitations under the License.
+ *
+ *
+ */
+
+package org.nd4j.linalg.learning;
+
+import lombok.Data;
+import org.apache.commons.math3.util.FastMath;
+import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.api.shape.Shape;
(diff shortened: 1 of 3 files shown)


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/NadamUpdater.java
```diff
@@ -0,0 +1,101 @@
...
+
+        INDArray oneMinusBeta1Grad = gradient.mul(1.0 - beta1);
+        m.muli(beta1).addi(oneMinusBeta1Grad);
+
+        INDArray oneMinusBeta2GradSquared = gradient.mul(gradient).muli(1.0 - beta2);
+        v.muli(beta2).addi(oneMinusBeta2GradSquared);
+
+        double beta1t = FastMath.pow(beta1, iteration + 1);
+
+        INDArray biasCorrectedEstimateOfMomentum = m.muli(beta1).divi(1.0 - beta1t);
```
AlexDBlack: I don't believe this is correct. This line is doing the following:
m_t = m_(t-1) * beta1 / (1-beta_1^t) then calling that 'biasCorrectedEstimateOfMomentum'. Mathematically, it's correct, except that you're incorrectly overwriting m_t in the process.

I believe it should be ```m.mul(beta1).divi(1.0 - beta1t);``` (i.e., mul, not muli - big difference here)

deil87: Thanks Alex!As I understand with `i` at the end we are mutating.  And we are going to use result of first mutation of m 
`m.muli(beta1).addi(oneMinusBeta1Grad);`
on  the next iteration, right? -> that is why we can use `.muli`. But `biasCorrectedEstimateOfMomentum` we need only on this current iteration and should make a copy..... and use `.mul` ?

AlexDBlack: Right, exactly. We've already calculated m_t from m_{t-1} and don't want to modify it again (it's needed next iteration).

In general: z = x.muli(y) -> x is modified, x and z are same object. Same thing for any x.opi(y), where "opi" is addi/muli/divi etc; add/mul/div etc (without the ```i``` at the end) are "copy first" ops.

Put another way:
```INDArray biasCorrectedEstimateOfMomentum = m.mul(beta1).divi(1.0 - beta1t);``` is equivalent to ```INDArray biasCorrectedEstimateOfMomentum = m.dup().muli(beta1).divi(1.0 - beta1t);```.

deil87:
@AlexDBlack Fixed this issue and added test to deeplearning4j repo https://github.com/deeplearning4j/deeplearning4j/pull/3583/commits/541d89576f7b0a5808ee7da4654dcec0f8a6d49e as well

AlexDBlack:
Merged, thanks!

deil87:
Thanks for cooperation!

Issue #1926, opened 2017-07-08
symmetricGeneralizedEigenvalues()


lukeczapla:
#### Issue Description

The functions in Eigen.java are mis-documented (it is indeed returning the Eigenvectors, which is what I want, and it is solving A x = L x) and I suggest these changes to Eigen.java JavaDoc for the `public static INDArray symmetricGeneralizedEigenvalues(INDArray A)`   :

https://gist.github.com/lukeczapla/3d79c65e1a8989721966bf6378843c1b

And here are my tests: https://gist.github.com/lukeczapla/f8e14e93dfa816a5ad88a2a466ff40c0

The other version `public static INDArray symmetricGeneralizedEigenvalues(INDArray A, INDArray B)` is not doing what it claims.  I could only think of something like this to do it, but I'm not sure if this is good:

`    /**
     * Compute generalized eigenvalues of the problem A x = L B x.
     *
     * @param A symmetric Matrix A.
     * @param B symmetric Matrix B.
     * @return a vector of eigenvalues L.
     */

    public static INDArray symmetricGeneralizedEigenvalues(INDArray A, INDArray B) {
        assert A.rows() == A.columns();
        assert B.rows() == B.columns();
        INDArray W = Nd4j.create(A.rows());
        Nd4j.getBlasWrapper().syev( 'V', 'L', InverseMatrix.inverse(B, false).mmul(A), W);
        return W;
    }
`

#### Version Information

ND4J 0.8.1-SNAPSHOT

#### Contributing

My code and documentation gist is above, and there is a related pull request that attempts to fix everything, keeping in mind that syev() is the lapack one that's best implemented at the moment.

saudet:
@rcorbish What do you think?

rcorbish:
I took a quick look. This looks fine, it's nice to have a better interface to syev in Eigen.

lukeczapla:
Thank you.  I have been doing some work on principal component analysis in my workspace, might this be useful for Nd4j?   The test is generating correlated data, performing an "inverse harmonic analysis", and then using this new symmmetricGeneralizedEigenvalues(A, B) to get the principal components along the variables (two files in one gist):

https://gist.github.com/lukeczapla/c9736b74e765478de277d68d3f8f8d40


You can see I'm still testing it but it's coming out correctly, putting the orthonormal basis vector of greatest variance first in the list (since syev() does that automatically with eigenvalues in increasing order).

saudet:
[...]

PR #1927, opened 2017-07-09
Eigen.java

lukeczapla:

In regards to issue #1926:

I tried to implement the suggestions on JavaDoc documenting the current symmetricGeneralizedEigenvectors(A) behavior.  That it is transforming matrix A into its eigenvectors in the process of returning the eigenvalues.  And solving the problem A x = L x.

and I added some code to symmetricGeneralizedEigenvectors(A, B) that should in theory solve the problem "A x = L B x" as it was intended.  I saw in some old branches that the code was entirely different.

*Updated: I made two versions of both methods, since the original symmetricGeneralizedEigenvalues(A) overwrote the matrix A with the eigenvectors as columns, I have one with an extra parameter boolean calculateVectors, if it's false, it doesn't do this.   Similarly with the two arg version symmetricGeneralizedEigenvalues(A, B), it was not saving the eigenvectors.  So I made a new version that has a third parameter calculateVectors, which if it's true, then it will assign to A in-place and collect eigenvectors as columns of A.   The idea behind the calculation is that if you are trying to solve "A x = L B x" and L is a scalar, then it's "A x = B L x" and the solution to B_inverse A x = L x (making substitution of A' = B_inverse A) should be the correct solution.  It's not efficient but it works with syev() call that is implemented *

## How was this patch tested?

https://gist.github.com/lukeczapla/f8e14e93dfa816a5ad88a2a466ff40c0

I did unit tests of the symmetricGeneralizedEigenvalues(A) vs. a version of eigenvector/eigenvalue calculator found in Numerical Recipes (Jacobi method).  The same eigenvector/value pairs are output but in sorted order by increasing eigenvalue from this syev() method.  I have tested all new versions with 6x6 symmetric matrix data and checked that the solutions of A x = L B x are coming out correctly, for the version with the two parameters A and B.

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/eigen/Eigen.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/eigen/Eigen.java
@@ -48,7 +48,7 @@ public class Eigen {
      * Compute generalized eigenvalues of the problem A x = L x.
      * Matrix A is modified in the process, holding eigenvectors at the end.
      *
-     * @param A symmetric Matrix A. After completion, A will contain the eigenvectors as rows
+     * @param A symmetric Matrix A. After completion, A will contain the eigenvectors as columns
      * @return a vector of eigenvalues L.
      */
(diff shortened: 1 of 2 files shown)


saudet:
I'd like to add @rcorbish as a reviewer but we can't :)

PR #1928, opened 2017-07-10
PCA updates and Transforms.mpow function

lukeczapla:

At the suggestion of @agibsonccc to go ahead and try to revamp the PCA routines, I tried to implement a PCA-type of method that takes data (records each with a fixed number of features) as rows (w/ features as columns) and calculates the covariance matrix, the mean values of each feature, the eigenvectors, and the eigenvalues, and then performs a dimensionality-reduction to a provided variance of the data.  Since it goes along with the existing PCA.java, I added it in there and started to think about everything.

The revamped PCA class has all the old static methods but can be instantiated as an object, and the reason is that since there is so much data reported, it helps to be able to access things like myPCA.getEigenvalues() and myPCA.getCovarianceMatrix().  In order to calculate the variance between data calculated with a limited basis set vs. the exact data represented in the full basis set, it was necessary to have those eigenvectors stored within the object.  The only other alternative was to write functions that return INDArray[4] which is not very readable and usable.

So the changes are meant to provide alternative usage of PCA and to maintain all existing routines.  To transform data from features into linear combinations of basis vectors, I multiply E_transpose*dx,  where E_transpose is the eigenvector matrix transposed (which is its inverse because it's orthonormal unit basis vector matrix) and dx is the difference between the features and their means for a given record.

## How was this patch tested?

TestPCA.java has some of my tests.  I have some other ones in a little app that I can post as a gist.

One of the important tests was that the dimensionality reduction indeed works to a given variance.  To do that, I calculated the projection of the data from the mean onto the basis vectors and then calculated the Euclidean distance to the data point without the removed dimensions vs. the same value with all dimensions (which exactly lies at the point of interest) and calculated a ratio.  I did it for 100 random samples and it seems to provide 87% of the variance when I told it to reduce it to within 70% variance.  A similar test that intends to reduce to 90% variance was actually able to reproduce 99.5% of the variance for 100 random samples.  So I am just erring on the side of being safe here, since when this thing is fed with 10000 data records that have 1000 features, it may not be as great as when fed with 10000 data records that have 6 features.   But since randn was used in the generation procedure and some variables were pre-correlated to others by addition & subtraction, I could also check by eye that the results make sense.

*I am definitely interested in someone providing large data (a lot of records each with many features) to compare the different methods here* 

I also addressed issue #1933 with a Transforms.mpow(INDArray, int n, boolean dup) to raise a square matrix to a positive, negative, or zero power.


lukeczapla:
I'm making a minor change, I was taking the sqrt(variance) so my estimates are more like 76% variance for the dimensionality reduction to 70%, and 99% variance for the dimensionality reduction to 90%.    Basically just squaring 0.87 and 0.995.  Pushed new minor update to PCA.java to remove Math.sqrt(of the sum of values squared) to just the sum of values squared.

saudet:
I didn't check the math, but other than that looks good to me! Thanks

lukeczapla:
screen shot 2017-07-10 at 9 52 41 pm

Yea for the math, this was the idea of the variance for each component in the reducedBasis method and the related static method.  But the variance of 100 random samples in TestPCA.java tests it directly, too.  If someone has real large data (w/ lots of features) I could give it a production test.

lukeczapla:
I think of the problem in terms of direction of force in physics.  The smallest eigenvalue is like the loosest spring that vibrates the most along the +-direction of its eigenvector (either in the direction or in the opposite direction of the vector from the mean), and the largest eigenvalue is like the stiffest spring that only vibrates a little along the +-direction of its eigenvector.  So if you have 16 features that picture extends into 16 dimensions.  But it turns out only 10 of those "modes" may be necessary to account for 97% of the variance in the data, because most of the fluctuations are in the direction of those vectors.   These modes do not lose any of the original features, but just couple them together into a reduced number of modes.

These things could literally be forces and you could use them to sample states that reproduce the same exact distribution if the data was purely bell-curved/gaussian.  It would basically produce fake data that appears to belong in the set.   I could add that to the code (sampling states using the Boltzmann distribution), @agibsonccc what do you think?  It's in my code already, which is (will be) open source, but I've been holding off on releasing it in Java form.

lukeczapla:
Dummy data generator has a big test in TestPCA.java, it reproduces the same quality of data (variance by feature, eigenvalues/vectors, and mean) as the actual input data records used to produce them, within the limits of your sample size.   10000 input data records (of 16 features) are analyzed and then 10000 dummy data records are created that then match the same principal components, eigenvalues, mean, and variances.

saudet:
@lukeczapla FWIW, I think you have enough tests in there :) Thanks

lukeczapla:
ok, I'm going to clean it up in terms of documentation and add a routine to convert your data into the component vectors about the mean, and then you'd drop a certain number of columns at the end.  They're nearly always the smallest numbers.  tensorMmul ftw :)

lukeczapla:
great, let me know if it's acceptable.  @agibsonccc said he's away until next week, I can do more work on it and maintain it.  I'm a high school teacher so I'm busy after the summer but I want people's feedback and can add more stuff as necessary.  right now everything's chill and I'm writing a lot of code, between Java and JavaScript.  So I'm back and forth and will come back and improve it as I get new ideas about how people would use this.

saudet:
Looks good to me, but let's try to get some feedback from @rcorbish. He's the last one to have updated code in that area. :) @rcorbish what do you think, good to merge?

PR #1947, opened 2017-07-19
<Do not Merge> Assert same shape for element wise ops

eraly:

Currently for a handful of element wise ops we only assert length. As a result I can do A.mul(B) where A is size (10,10) and B is size(2,5,10) . 

## How was this patch tested?
Simple unit test for mul. But assert was added and checked for all element wise ops.

Please review


eraly:
@huitseeker On the right track. Possibly. 
I ran dl4j-tests and a small subset of tests break which will pass with a minor reshape:
https://github.com/deeplearning4j/deeplearning4j/compare/earlybird

@raver119 says it might be by design so not merging for now...

eraly:
Merge https://github.com/deeplearning4j/deeplearning4j/pull/3772 after this is merged

Issue #1950, opened 2017-07-20
INDArray.get wrong behavior


t-li:
#### Issue Description
In Scala with Nd4j.0.8.0 CUDA
val a = Nd4j.zeros(4,3,3)
a.get(point(0), interval(1,3), interval(1,3)) addi 1.0
println(a)

// yield wrong result
[[[0.00, 0.00, 1.00],
  [1.00, 0.00, 1.00],
  [1.00, 0.00, 0.00]],

 [[0.00, 0.00, 0.00],
  [0.00, 0.00, 0.00],
  [0.00, 0.00, 0.00]],

 [[0.00, 0.00, 0.00],
  [0.00, 0.00, 0.00],
  [0.00, 0.00, 0.00]],

 [[0.00, 0.00, 0.00],
  [0.00, 0.00, 0.00],
  [0.00, 0.00, 0.00]]]

// reset back to all zeros
a muli 0.0

// redo it in a different way
a.get(point(0), interval(1,3), point(1)) addi 1.0
a.get(point(0), interval(1,3), point(2)) addi 1.0
println(a)

// yield correct result
[[[0.00, 0.00, 0.00],
  [0.00, 1.00, 1.00],
  [0.00, 1.00, 1.00]],

 [[0.00, 0.00, 0.00],
  [0.00, 0.00, 0.00],
  [0.00, 0.00, 0.00]],

 [[0.00, 0.00, 0.00],
  [0.00, 0.00, 0.00],
  [0.00, 0.00, 0.00]],

 [[0.00, 0.00, 0.00],
  [0.00, 0.00, 0.00],
  [0.00, 0.00, 0.00]]]

#### Version Information
ND4J CUDA 0.8.0
CUDA 0.8.0
Nvidia driver 375.39
OS Ubuntu 14.04

PR #1951, opened 2017-07-20
More tests for indexing issues

eraly:
testGet added to c ordering
Fails with both f and c
Also seeing strange behaviour with get when the view subsetted is rank = 3 ...
Possibly related to this 
https://github.com/deeplearning4j/nd4j/issues/1950


agibsonccc:
Mind fixing merge conflicts?

PR #1976, opened 2017-07-25
Minor cnpy fixes

raver119:
Minor cnpy fixes

- CUDA works
- Memory gets released if numpy array was originated from file.
- we are ***NOT*** releasing original pointer, if NDArray was created from numpy pointer. Numpy should handle that, so we won't get into double spending crash.


raver119:
cc @agibsonccc

PR #2000, opened 2017-08-01
Configure surefire for nd4j-jackson-reflectionloader to avoid failure in maven tests

AlexDBlack:
Fixes tests in nd4j-jackson-reflectionloader incorrectly failing due to maven surefire config. These were already passing in IntelliJ


huitseeker:
LGTM

PR #2007, opened 2017-08-02
Add a script for tagging and making the release

huitseeker:


huitseeker:
Probably obsolete given the SKIPBUILD option in the release, @saudet ?

saudet:
Ah, yes, let's use perform-release.sh?

PR #2016, opened 2017-08-04
Version Check

AlexDBlack:
This PR: adds version check functionality, based on git-commit-id plugin.
Checks for:
- Mixed DL4J/ND4J/DataVec etc versions
- Mixed Scala versions, our deps only (i.e., dl4j spark 2.10, datavec spark 2.11)
- Mixed spark versions (i.e., datavec spark 1, dl4j spark 2)

As far as I can tell: this is complete + good to go. Can be merged after final review.

EDIT: One issue discovered here with IntelliJ (I've been unable to find a workaround after many hours) - it doesn't  work with Intellij running with "dynamic.classpath" in workspace.xml set to true.
It seems the reflections library simply fails to find any -git.properties files, even though they definitely are on the classpath (and can be found using a hard-coded path using say ClassPathResource).
The exact same code/pom runs perfectly with intellij not using dynamic classpath, or when running via an uber-jar.
Workaround for now: if no git.properties files are found, we abort the version check.

----
Unfortunately it does not seem possible to automate testing of this functionality - I could not find a way to manually control the classpath for individual tests for Maven Surefire.

Output of VersionCheck.logVersionInfo();
https://gist.github.com/AlexDBlack/1c59e94be4c79c597b01206995502aae

Output of VersionCheck.logVersionInfo(VersionCheck.Detail.FULL);
https://gist.github.com/AlexDBlack/730075b856f74ea8002b2e78a75c7e2f

Output when forcing datavec-api 0.9.0 via dependency management:
https://gist.github.com/AlexDBlack/46ebc7a574946a33d9872e7faf73760c

Output when nd4j-jblas or Canova are on the classpath (JVM exits immediately)
https://gist.github.com/AlexDBlack/abfb9b042fdbdbf70a871bb9f1b54ddf

Output with datavec spark 2.10 and 2.11:
https://gist.github.com/AlexDBlack/1b410f575db8d242669db6c4c1a42835

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/versioncheck/VersionInfo.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/versioncheck/VersionInfo.java
@@ -15,7 +15,7 @@ import java.util.Properties;
 @NoArgsConstructor
 @Data
 @Builder
-public class GitRepositoryState {
+public class VersionInfo {
 
     private static final int SUFFIX_LENGTH = "-git.properties".length();
 
@@ -45,13 +45,13 @@ public class GitRepositoryState {
     private String buildHost;               // =${git.build.host}
     private String buildVersion;             // =${git.build.version}
 
-    public GitRepositoryState(String groupId, String artifactId, String buildVersion){
+    public VersionInfo(String groupId, String artifactId, String buildVersion){
         this.groupId = groupId;
         this.artifactId = artifactId;
         this.buildVersion = buildVersion;
     }
 
-    public GitRepositoryState(String propertiesFilePath) throws IOException {
+    public VersionInfo(String propertiesFilePath) throws IOException {
         //First: parse the properties file path, which is in format <groupid>-<artifactId>-git.properties
         int idxOf = propertiesFilePath.lastIndexOf('/');
         idxOf = Math.max(idxOf, propertiesFilePath.lastIndexOf('\\'));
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -684,7 +687,7 @@
             <dependency>
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
-                <version>18.0</version>
+                <version>20.0</version>
```
agibsonccc: Maybe give guava a ${guava.version} ?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/versioncheck/VersionInfo.java
```diff
@@ -0,0 +1,97 @@
...
+    private String artifactId;
+
+    private String tags;                    // =${git.tags} // comma separated tag names
+    private String branch;                  // =${git.branch}
+    private String dirty;                   // =${git.dirty}
+    private String remoteOriginUrl;         // =${git.remote.origin.url}
+
+    private String commitId;                // =${git.commit.id.full} OR ${git.commit.id}
+    private String commitIdAbbrev;          // =${git.commit.id.abbrev}
+    private String describe;                // =${git.commit.id.describe}
```
AlexDBlack: I'm OK with other values being absent - as long as the main ones (groupid, artifact, version, commit id) are present.

pom.xml
AlexDBlack: Done. I also cleaned up the maven plugin versions (moved to properties also).

pom.xml
```diff
@@ -428,9 +428,11 @@
         <maven-build-helper-plugin.version>1.12</maven-build-helper-plugin.version>
```
huitseeker: While you're at it, can you bump this to 3.0.0 ? (it's what we use throughout the rest of the code base).

PR #2021, opened 2017-08-05
Introduce SKIP_BUILD mode to perform-release.sh script

saudet:

Allow updating the source code for a release before making the final build on the "latest_release" tag.

## How was this patch tested?

Build succeeds with source code modified by the script, with no reference to SNAPSHOT artifacts.

### DIFF
--- a/perform-release.sh
+++ b/perform-release.sh
@@ -9,9 +9,7 @@ fi
 RELEASE_VERSION=$1
 SNAPSHOT_VERSION=$2
 STAGING_REPOSITORY=${3:-}
-if [[ -z ${SKIP_BUILD:-} ]]; then
-    SKIP_BUILD=0
-fi
+SKIP_BUILD=${SKIP_BUILD:-0}
 
 echo "Releasing version $RELEASE_VERSION ($SNAPSHOT_VERSION) to repository $STAGING_REPOSITORY"
 echo "========================================================================================"
--- a/perform-release.sh
+++ b/perform-release.sh
@@ -28,8 +28,8 @@ if [[ -z $(git tag -l "libnd4j-$RELEASE_VERSION") ]]; then
         bash buildnativeoperations.sh -c cuda -v 7.5
         bash buildnativeoperations.sh -c cuda -v 8.0
     fi
-    git tag -a -m "libnd4j-$RELEASE_VERSION" "libnd4j-$RELEASE_VERSION"
-    git tag -a -f -m "libnd4j-$RELEASE_VERSION" "latest_release"
+    git tag -s -a -m "libnd4j-$RELEASE_VERSION" "libnd4j-$RELEASE_VERSION"
+    git tag -s -a -f -m "libnd4j-$RELEASE_VERSION" "latest_release"
 fi
 cd ../nd4j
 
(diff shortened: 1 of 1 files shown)


perform-release.sh
```diff
@@ -9,6 +9,9 @@ fi
 RELEASE_VERSION=$1
 SNAPSHOT_VERSION=$2
 STAGING_REPOSITORY=${3:-}
+if [[ -z ${SKIP_BUILD:-} ]]; then
```
huitseeker: Here, you can just `SKIP_BUILD=${SKIP_BUILD:-0}`

perform-release.sh
```diff
@@ -22,58 +25,65 @@ cd ../libnd4j
...
-if [[ -z ${STAGING_REPOSITORY:-} ]]; then
-    # create new staging repository with everything in it
-    source change-scala-versions.sh 2.10
-    source change-cuda-versions.sh 7.5
-    mvn clean deploy -Dgpg.executable=gpg2 -DperformRelease -Psonatype-oss-release -DskipTests -Denforcer.skip -DstagingRepositoryId=$STAGING_REPOSITORY
-
-    if [[ ! $(grep stagingRepository.id target/nexus-staging/staging/*.properties) =~ ^stagingRepository.id=(.*) ]]; then
-        exit 1
+if [[ "${SKIP_BUILD}" == "0" ]]; then
+    if [[ -z ${STAGING_REPOSITORY:-} ]]; then
```
huitseeker: So, this reproduces the logic of buildmultipleScalaVersions.sh, except imperfectly (where is the cuda 7.5 / Scala 2.10 or cuda 8.0 / Scala 2.11 .. that some of this works is a happy artefact that the only thing we need for Scala is Cuda-independent  — parameter server and kryo).

How about reusing parts of `buildmultiplescalaversions.sh` ? And  taking inspiration from the logic in `buildAllversions.sh` ? (where both cuda versions are addressed)

perform-release.sh
```diff
@@ -22,58 +25,65 @@ cd ../libnd4j
...
-    bash buildnativeoperations.sh -c cpu
-    bash buildnativeoperations.sh -c cuda -v 7.5
-    bash buildnativeoperations.sh -c cuda -v 8.0
+    if [[ "${SKIP_BUILD}" == "0" ]]; then
+        bash buildnativeoperations.sh -c cpu
+        bash buildnativeoperations.sh -c cuda -v 7.5
+        bash buildnativeoperations.sh -c cuda -v 8.0
+    fi
     git tag -a -m "libnd4j-$RELEASE_VERSION" "libnd4j-$RELEASE_VERSION"
+    git tag -a -f -m "libnd4j-$RELEASE_VERSION" "latest_release"
```
huitseeker: For operations fo this level of importance on the repo I wouldn't mind a `-s`, if possible.

saudet: Sure, I suppose. Should we use it for all tags related to releases?

huitseeker: I use `-s` in all my git operations, but this is controversial : committers not equipped with the ability to sign commits will balk at the inconvenience.

Issue #1991, opened 2017-07-29
Problems converting Java.ArrayList<Float> to NDArray in scala.


borissmidt:
#### Issue Description
I'm trying to convert a Java.util.List<java.Float> into a NDArray:
I have the following scala code:
```
val vector: util.List[Float] = model.getVector("word")
val scalaArray = vector.asScala.toArray
```
Then if i try to convert it, with `scalaArray.toNDArray()` i with the implicit function it doesn't find the implicit conversion:
Then i tried then i tells me that it cannot convert java.Float to an NDArray, since this is a java object But scala has no real way to convert it explicitly to an:
```
floatArray2INDArray(scalaArray).toNDArray()
```
and even this:
```
floatArray2INDArray(scalaArray.map(_.toFloat)).toNDArray()
```
Eventually i just solved it in java code like this:
```
    public INDArray getWordVector(String word){
        float[] buffer = new float[model.getDim()];
        List<Float> vector = model.getVector(word);
        for(int i = 0;i < model.getDim() ;i++){
            buffer[i] = vector.get(i);
        }
        return Nd4j.create(buffer);
    }
```
a scala oneliner solution is where to option allows you to unbox the Float object into a primitive float:
```
    val scalaArray = vector.asScala.toArray.flatMap(float => Option(float).map(_.toFloat)).toNDArray
```
- expected behavior
Direct conversion functions between Seq/ List containers To NDArrays. Since most java/ scala code operate on List/Seq and not the Array.

- encountered behavior

It is hard to convert from a boxed java.Float to a scala.Float in scala. 
There are no functions in java to from a java.List to Float[] without manually copying.

I understand that this is a problem on Scala its side because opaque boxing and unboxing of a Float.
On the java side it is the missing functions to convert between Java List to Array and functions to go from java Float[] to float[]

#### Version Information

Please indicate relevant versions, including, if relevant:

* Deeplearning4j version
ND4j 0.8.0
ND4s 0.8.0
#### Contributing

If you'd like to help us fix the issue by contributing some code, but would
like guidance or help in doing so, please mention it!
Since i could write convenience code to do the conversions from the scala types and the java list. But these probably have to be documented that they are rather inefficient because of the way they work.

saudet:
That's a general limitation in Java as well. Since we need to copy the values from arrays anyway, the equivalent with List wouldn't be that inefficient. Additional factory methods are welcome! Please send your pull request. :)

borissmidt:
[...]

PR #2043, opened 2017-08-12
added conversion from List<Float> and List<Double> to NDarray

borissmidt:
#1991

## How was this patch tested?

I tested in manually using the scala repl. 
Please review

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4j.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4j.java
@@ -20,6 +20,8 @@
 package org.nd4j.linalg.factory;
 
 import com.google.common.base.Function;
+import com.google.common.primitives.Doubles;
+import com.google.common.primitives.Floats;
 import com.google.common.primitives.Ints;
 import lombok.NonNull;
 import org.apache.commons.io.FileUtils;
@@ -3268,6 +3270,28 @@ public class Nd4j {
      */
     public static INDArray create(float[] data) {
         return create(data, order());
+    }
+
+
+    /**
+     * Creates a row vector with the data
+     *
+     * @param list the columns of the ndarray
+     * @return the created ndarray
+     */
+    public static INDArray create(List<? extends Number> list) {
+        INDArray array = create(list.size());
+        int cnt = 0;
+        if (dataType() == DataBuffer.Type.DOUBLE) {
+            for (Number element: list) {
+                array.putScalar(cnt++,element.doubleValue());
+            }
+        } else {
+            for (Number element : list) {
+                array.putScalar(cnt++,element.floatValue());
+            }
+        }
+        return array;
     }
 
     /**
--- a/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/factory/Nd4jTest.java
+++ b/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/factory/Nd4jTest.java
@@ -7,6 +7,9 @@ import org.nd4j.linalg.BaseNd4jTest;
 import org.nd4j.linalg.api.ndarray.INDArray;
 import org.nd4j.linalg.api.rng.Random;
 
+import java.util.Arrays;
+import java.util.List;
+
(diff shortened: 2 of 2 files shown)


saudet:
I think it would be clearer to overload the Nd4j.create() methods. What do you think @huitseeker?

borissmidt:
@saudet i would really liked to do that, but because of the type erasure in java, f(List<A> x) is seen as the same as f(List<B> x). 
I'm not adept enough in java to solve this problem, so if you can assist i would be happy to help and probably learn something new :).

saudet:
Ah yes that's an issue. :) LGTM then

raver119:
Why not use single overload List<? extends Number>? We have fairly limited number of datatypes, and any Number should be converted appropriately?

borissmidt:
Yes it seems like a good idea, How would i know if i have to call `Floats.toArray` or `Doubles.toArray` should i check the instance of the first element of the List?
```
public static INDArray createFloat(List<? extends Number>[] data) {
Number element = data.get(0)
if( element instanceof Float)
else if (element instanceof Double)
else if (element instanceof Long)
else if (element instanceof Integer)
}
        
```

raver119:
No. You don't do that. Number has some methods, lint floatValue, doubleValue etc. So you just check for Nd4j.dataType(), and then use the same method (i.e. floatValue) on all elements to fill NDArray

raver119:
I mean - it doesn't matters what exactly you have as Number. The only thing that matters is current datatype of Nd4j. Once you know that - you just use matching method of Number class. i.e. floatValue()

borissmidt:
```
    public static DataBuffer createBuffer(double[] data) {
        DataBuffer ret;
        if (dataType() == DataBuffer.Type.DOUBLE)
            ret = Nd4j.getMemoryManager().getCurrentWorkspace() == null ? DATA_BUFFER_FACTORY_INSTANCE.createDouble(data) : DATA_BUFFER_FACTORY_INSTANCE.createDouble(data, Nd4j.getMemoryManager().getCurrentWorkspace());
        else if (dataType() == DataBuffer.Type.HALF)
            ret = Nd4j.getMemoryManager().getCurrentWorkspace() == null ? DATA_BUFFER_FACTORY_INSTANCE.createHalf(data) : DATA_BUFFER_FACTORY_INSTANCE.createHalf(ArrayUtil.toFloats(data), Nd4j.getMemoryManager().getCurrentWorkspace());
        else
            ret = Nd4j.getMemoryManager().getCurrentWorkspace() == null ? DATA_BUFFER_FACTORY_INSTANCE.createFloat(ArrayUtil.toFloats(data)) : DATA_BUFFER_FACTORY_INSTANCE.createFloat(ArrayUtil.toFloats(data), Nd4j.getMemoryManager().getCurrentWorkspace());
        logCreationIfNecessary(ret);
        return ret;
    }
```
So more like this function? So the usage of a halve/float/double array is an external setting?

raver119:
nah. you want to create NDArray, not a buffer.

So something like this (beware, pseudo code!):
```
public static <T extends Number> INDArray create(List<T> list) {
   if (Nd4.dataType() == DataBuffer.Type.DOUBLE) { 
   double[] tempArray = new double[list.size()];
   int cnt = 0;
   for (T element: list) {
     tempArray[cnt++] = element.doubleValue();
  }
 return Nd4j.create(tempArray);
 } else {
 float[] tempArray = new float[list.size()];
   int cnt = 0;
   for (T element: list) {
     tempArray[cnt++] = element.floatValue();
  }
 return Nd4j.create(tempArray);
 }
}
```

So, in this case i'm always using double as temporary array, If current dataType is double. And using float temp array in all other cases.

Obviously, the same could be written without temp array and no code branching, just create empty NDArray of desired length, and call putScalar in loop.

raver119:
Hm. Code turned out to be not a pseudo code :/

borissmidt:
I will try to make the implementation with the putScalar in the loop but would it really be more efficient?

raver119:
It might be a bit slower (neglectable though), but more memory-friendly.

borissmidt:
I will add them this weekend, when i have the time to do it.

huitseeker:
@borissmidt Any news for us? Do you think you'll have time to push this one over the line soon? It's so close!

borissmidt:
I'm on vacation this is a why i am unable to do it now.

PR #2093, opened 2017-09-02
Minor printing improvements esp if column vector

eraly:
https://github.com/deeplearning4j/nd4j/issues/1929

## How was this patch tested?
Unit test

Please review

### DIFF
--- a/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/imports/TFGraphTestAllHelper.java
+++ b/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/imports/TFGraphTestAllHelper.java
@@ -89,7 +89,7 @@ public class TFGraphTestAllHelper {
             for (String outputNode : predictions.keySet()) {
                 INDArray nd4jPred = graph.getVariable(outputNode).getArr();
                 INDArray tfPred = predictions.get(outputNode);
-                assertEquals("Predictions do not match on " + modelName, tfPred.reshape(nd4jPred.shape()), nd4jPred);
+                assertEquals("Predictions do not match on " + modelName, tfPred, nd4jPred);
             }
             log.info("\n\tTEST " + modelName + " PASSED...");
             log.info("\n========================================================\n");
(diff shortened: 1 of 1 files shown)


agibsonccc:
Got a sample of the output?

eraly:
This is 5x1:

[ 1.00, 
  2.00, 
  3.00, 
  4.00, 
  5.00]

This is 2x4x1:

[[ 1.00, 
  3.00, 
  5.00, 
  7.00],  

 [ 2.00, 
  4.00, 
  6.00, 
  8.00]]

AlexDBlack:
Does this mean that a [1000,1] will print as 1000 lines, instead of the current 1 line? If so, I'm not sure how much I like that...
an alternative might be to use a different delimiter, like ";"?
i.e., a 5x1 would print as ```[1; 2; 3; 4; 5]```

agibsonccc:
That's actually more of what people expect though. I would be ok with this.
We can control the truncation remember.

eraly:
@AlexDBlack Yes. Currently it will print 1000 lines.
I can make some changes if we want to truncate the printed string.
Say if it's bigger than say 150 (?) we just do "..." and the last row?

AlexDBlack:
Limiting here makes sense, and first N, ..., last is good.
As for the specific limit - we can make it reasonably generous, maybe 500 or so?

eraly:
So I had this nagging feeling this was going to break something that I had written a long time ago and I couldn't put my finger on it. And now I know what:

https://github.com/deeplearning4j/nd4j/blob/88f62abd5b43489942f4394bf90f901de5b09a17/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4j.java#L2401

eraly:
I need to make sure that what I am doing here doesn't break the NdArrayJSONReader and NdArrayJSONWriter or the readTxt/writeTxt functionality in Nd4j.

The tests for them don't cover the column vector edge case.  We do want to support the JSON reader/writer functionality, right? It might be correct but it's pretty brittle (I wrote it a long time ago so you can blame me) and there is plenty of cleaning up I can do there...

AlexDBlack:
Quick thought: I wonder if we can leverage (or share code with) some of the existing Jackson serializers for JSON?

eraly:
I will look into leveraging what we have in jackson for this.

eraly:
I will get back to this after the use case disco

huitseeker:
@eraly Any news on getting back to this one?

eraly:
Hang up was what to do with the NdArrayJSONReader and NdArrayJSONWriter. Tests for these are ignored in current master and I think in the past release as well. I am voting to deprecate it. I don't see it being used anywhere.

huitseeker:
@eraly I <3 deprecations! Go ahead!

agibsonccc:
@eraly we should probably just move forward with deletions here.  We're doing a major version bump anyways.

eraly:
Overrode this branch by mistake TF import stuff :( Will reopen

PR #2116, opened 2017-09-10
Add basic momentum optimizer

firasdib:
This PR adds the rudimentary momentum optimizer to the DL4J repertoar of optimizers.
The implementation is straight forward; basically a stripped down version of the nesterovs optimizer.

It is not uncommon to see these two combined into one class/function call, with a parameter defining `nesterovs = true/false`.

Either way, this seems to work for me and figured I'd share it with you.

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/MomentumUpdater.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/MomentumUpdater.java
@@ -1,0 +1,59 @@
+package org.nd4j.linalg.learning;
+
+import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.api.shape.Shape;
+import org.nd4j.linalg.learning.config.Momentum;
+
+public class MomentumUpdater implements GradientUpdater<Momentum> {
+    private final Momentum config;
+    private INDArray v;
+    private char gradientReshapeOrder;
+
+    public MomentumUpdater(final Momentum config) {
+        this.config = config;
+    }
+
+    @Override
+    public Momentum getConfig() {
+        return config;
+    }
+
+    @Override
+    public void setStateViewArray(final INDArray viewArray, final int[] gradientShape, final char gradientOrder,
+                                  final boolean initialize) {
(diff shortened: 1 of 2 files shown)


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/MomentumUpdater.java
```diff
@@ -0,0 +1,59 @@
...
+        if (v == null) {
+            throw new IllegalStateException("Updater has not been initialized with view state");
+        }
+
+        final double momentum = config.getMomentum();
+        final double learningRate = config.getLearningRate();
+
+        // Standard momentum: v = mu * v_prev - learning_rate * gradient(x)
+        final INDArray vPrev = v.dup(gradientReshapeOrder);
+        vPrev.muli(momentum).subi(gradient.dup(gradientReshapeOrder).muli(learningRate));
```
AlexDBlack: This is incorrect. You dup, but never actually update the original ```v``` array - it will *always* be equal to the initial value (0), and consequently equal to vanilla SGD in practice.
Instead, it should be ```v.muli(momentum).subi(gradient.dup(order).muli(learningRate))```

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/learning/MomentumUpdater.java
```diff
@@ -0,0 +1,59 @@
...
+
+        gradient.addi(vPrev);
```
AlexDBlack: Should be ```gradient.assign(v)``` not addi - gradient updating is in-place, and the update is equal to ```v```

firasdib:
@agibsonccc I'm not sure this should have been merged, its not compatible with @AlexDBlack's latest changes with regards to learning rate schedules and similar. It's a quick fix but I'm currently swamped I'm afraid.

agibsonccc:
Will revert. From what I understood with alex, we should have only consolidated all of the learning rates to work with updaters only.

AlexDBlack:
Yeah, just needs to be updated - PR/review was just before those changes.

PR #2161, opened 2017-10-01
Upgrade to CUDA 9.0

saudet:
WORK IN PROGRESS: DO NOT MERGE.

Haven't had time to test it out, but it should work. I'll be releasing the presets once we've confirmed that there are no issues. And I'll put back CUDA 8.0 as default before merging.

Note: This also drops support for CUDA 7.5.


saudet:
I've made a release version of the presets, and everything seems to be working fine on my end, so this is good to merge!

Though, should we leave it with CUDA 9.0 as default? Or should we switch the default back to CUDA 8.0? I'll leave it up to whoever merges this.

raver119:
Right now we support 7.5 and 8.0. 8.0 is default. I think it would make sense to have latest version as default.

awsfanatic:
Shouldn't we also update https://github.com/deeplearning4j/nd4j/blob/master/buildAllversions.sh?

raver119:
good catch.

ср, 4 окт. 2017 г. в 23:34, Sergey Kovalev <notifications@github.com>:

PR #2181, opened 2017-10-10
Update Nd4jCpuPresets to map new native operations

saudet:

Update `Nd4jCpuPresets.java` to map new functions up to "ops/declarable/CustomOperations.h".

## How was this patch tested?

It builds fine on both Linux and Windows with https://github.com/deeplearning4j/libnd4j/pull/505, but new functionality has not been tested.


nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/nativeblas/Nd4jCpuPresets.java
```diff
@@ -26,23 +26,149 @@
...
+                "nd4j::ops::maxpool2d",
+                "nd4j::ops::avgpool2d",
+                "nd4j::ops::pnormpool2d",
+                "nd4j::ops::maxpool3d_bp",
+                "nd4j::ops::avgpool3d",
+                "nd4j::ops::avgpool3d_bp",
+                "nd4j::ops::fullconv3d",
+                "nd4j::ops::fullconv3d_bp",
+                "nd4j::ops::fullconv3d_grad",
+                "nd4j::ops::maxpool2d_bp",
```
raver119: Is there any way we could provide auto-updates for this list? Some script maybe?

saudet: A script that would be part of libnd4j? We could read something from a file somewhere too, it's just Java after all...

saudet:
Ok, I've updated the presets, and now the ops get picked up automatically from CustomOperations.h and they get listed as Java Class objects in `Nd4jCpu.floatOps`, `Nd4jCpu.halfOps`, and `Nd4jCpu.doubleOps`. Looks good? Let me know if there is anything else that should be added/changed!

raver119:
@agibsonccc do you want something else for SameDiff?

agibsonccc:
Yes that should work as a start..we'll still need to do something about generating class declarations for the classes though.

saudet:
Which set of classes?

agibsonccc:
@saudet Similar to: https://github.com/deeplearning4j/nd4j/blob/2b339a83f39d160169df9ed14fb54c7652970a19/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/layers/convolution/Conv2D.java#L25  mapping the c++ equivalent.

saudet:
It's there, float_conv2d is here, for example:
https://github.com/deeplearning4j/nd4j/blob/36fadf7e5b7a8efddf30187125ec6430bb0bb8f5/nd4j-backends/nd4j-backend-impls/nd4j-native/src/main/java/org/nd4j/nativeblas/Nd4jCpu.java#L12227
Don't ask me how we're supposed to use it though :)

agibsonccc:
I'll let this go for now. We'll figure out the rest later..I think the code gen we want to do is going to be substantially more complicated than is worth it right now.

saudet:
I see what you mean. What would help is to agree on a set of Java interfaces to implement, which could realistically be implemented in C++ using types from there, like org.nd4j.nativeblas.NativeOps, and then we could have the C++ classes automatically implement them.

agibsonccc:
Right so let's merge this for now and then do a separate PR with more details. I agree with you about the set of common interfaces etc. We m ostly have that defined.

PR #2182, opened 2017-10-11
Specifying the value of a property with a name that conflicts with an environment variable is not a default, it's an override.

huitseeker:
This ignored any libnd4j directory that isn't an immediate sibling
of the nd4j checkout (breaking, e.g. native packaging, or any custom
build).

it does mean we'll have to set the LIBND4J_HOME variable.

### DIFF
--- a/nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom.xml
+++ b/nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom.xml
@@ -12,7 +12,6 @@
     <properties>
         <cuda.version>9.0</cuda.version>
         <cudnn.version>7.0</cudnn.version>
-        <env.LIBND4J_HOME>${basedir}/../../../../libnd4j/</env.LIBND4J_HOME>
     </properties>
 
     <build>
--- a/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml
+++ b/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml
@@ -12,7 +12,6 @@
     <properties>
         <openblas.version>0.2.19</openblas.version>
         <mkl.version>2017.3</mkl.version>
-        <env.LIBND4J_HOME>${basedir}/../../../../libnd4j/</env.LIBND4J_HOME>
     </properties>
 
     <dependencies>


huitseeker:
It's there already : 
https://github.com/huitseeker/nd4j/blob/1172232719bb2d26192807048ddf1c1b685907bc/nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom.xml#L146
https://github.com/huitseeker/nd4j/blob/1172232719bb2d26192807048ddf1c1b685907bc/nd4j-backends/nd4j-backend-impls/nd4j-native/pom.xml#L206

saudet:
We could use a profile?

When do we need to put libnd4j and nd4j in different directories anyway?

huitseeker:
1. We could use a non-default profile that you're encouraged to introduce for developer convenience in a later PR, @saudet. Mechanisms that set a value which looks at the LIBND4J_HOME environment variable and provides a default if it's unset are also fine. 
2. Libnd4j and nd4j are always in different directories, and that's fine. This is about the removed line forcing it to specifically be a sibling directory of nd4j. We need libnd4j assets to be in another directory every time we need nd4j's compilation checkout to be somewhere not at the same place as where libndj's build was produced. That's true of most *non-developer* builds, including every build pattern that cares about capturing the results of a build stage and moving them around, specially *most CI situations* and *every* native packaging situation.  The current setup would violate most of the java library packaging guidelines of Linux distros, including all major ones.
3. Also:
- this overrides an environment variable our doc takes painful steps to have users set carefully, 
- creates a hard to debug error, and 
- generally is the classic antipattern of hardcoding a path.

saudet:
I understand that this is a bug, but I don't understand why we have to
force users to set a environment variable when we can set a sensible
default! You're trying to make things harder to use than they need to be,
I'm sorry, I just don't agree, but if it's fine with everyone, I'm not
going to care more than pointing this out.

agibsonccc:
@saudett @huitseeker it's a tough call to make here. I would rather have a bit of friction and easier to debug builds. Beyond that, we can automate that LIBND4J_HOME env variable with the build-dl4j-stack.sh - if we can make it robust enough it won't matter anyways.

PR #2194, opened 2017-10-18
Swish

raver119:
Swish activation wrappers, java side

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/transforms/SwishDerivative.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/transforms/SwishDerivative.java
@@ -1,0 +1,169 @@
+/*-
+ *
+ *  * Copyright 2015 Skymind,Inc.
+ *  *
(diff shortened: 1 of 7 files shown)


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/transforms/SwishDerivative.java
```diff
@@ -0,0 +1,169 @@
...
+                            xAlongDimension.length());
+        else
+            return new SwishDerivative(x.tensorAlongDimension(index, dimension),
+                            z.tensorAlongDimension(index, dimension), xAlongDimension.length());
+
+    }
+
+    private static double swishDeriv(double input) {
+        double ex = FastMath.pow(Math.E, input);
+        return (ex * (input + ex + 1)) / FastMath.pow((ex + 1) , 2);
```
AlexDBlack: As per libnd4j PR: I'm not following this. It *might* be correct, but I can't see it...

raver119: Merging as discussed in gitter

PR #2213, opened 2017-10-25
INDArray: Modify incorrect javadoc

not-for-me:

This commit change some incorrect parameter descriptions in INDArray.

## How was this patch tested?
This change doesn't need to be tested.


AlexDBlack:
Thanks!

PR #2228, opened 2017-10-30
add new tools

clavvis:

These are new tools for dl4j users. Tools show several class content and show other informations.
User can save informations in file. Later I can add other functions.

Later I'd like to add some dl example using these tools.

## How was this patch tested?

Includes unit tests.


nd4j-common/pom.xml
```diff
@@ -53,6 +53,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.deeplearning4j</groupId>
+            <artifactId>deeplearning4j-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
raver119: That's not good. nd4j should NOT rely on dl4j.

nd4j-common/src/main/java/org/nd4j/tools/ShowDataExample.java
```diff
@@ -0,0 +1,63 @@
...
+		oinfo = "";
+		oinfo += BTools.getMtLvESS( mtLv );
+		oinfo += methodName + ": ";
+		sis.info( oinfo );
+		//
+        int minibatchSize = 128;
+        int rngSeed       = 12345;
+		//
+        //MNIST data for training
+        DataSetIterator train_DSI = new MnistDataSetIterator( minibatchSize, true, rngSeed );
```
raver119: Why do we have MNIST iterator in nd4j?

raver119:
Just one question: why this comes to nd4j?

agibsonccc:
@clavvis the utilities (not examples) are the only things that should go in here. @raver119 I told him to put common utility classes in nd4j-common.

clavvis:
Problem is caused by my example. I delete it from here. I can make it in dl4j-examples.

agibsonccc:
Yeah move that to dl4j-examples and get rid of the dl4j- dependency.

clavvis:
I deleted example and clear pom.xml. Later I will put example to dl4j-examples.

clavvis:
But because of dependency problem could be better create for universal tools stand-alone module in highest level (there is e.g. deeplearning4j, nd4j, ...).
For example I prepare report (for DL4JTools) showing content of MultiLayerNetwork and Layers. Now it is dependency problem.

nd4j-common/src/test/java/org/nd4j/tools2/DL4JToolsTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+public class DL4JToolsTest {
+	//
+	@Rule
+	public TemporaryFolder tmpFld = new TemporaryFolder();
+	//
+	private SIS sis;
+	//
+	@Test
+	public void test_all() throws Exception {
```
agibsonccc: Mind making the tests camel case?

nd4j-common/src/test/java/org/nd4j/tools/SISTest.java
```diff
@@ -0,0 +1,59 @@
...
+public class SISTest {
+	//
+	@Rule
+	public TemporaryFolder tmpFld = new TemporaryFolder();
+	//
+	private SIS sis;
+	//
+	
+	@Test
+	public void test_all() throws Exception {
```
agibsonccc: Camel case.

nd4j-common/src/test/java/org/nd4j/tools2/DL4JToolsTest.java
clavvis: I'm sorry, my English is bad. I found camel case here: https://en.wikipedia.org/wiki/Camel_case.
Do you mean change test_all to testAll? It is no problem.

agibsonccc:
Yep!

clavvis:
I made change to "testAll".

agibsonccc:
Can you change this for every test with an under score?

clavvis:
Instead of "test_getMtLvESS" you prefer "testgetMtLvESS" or "testGetMtLvESS"?

agibsonccc:
@clavvis yes that's right thanks!

clavvis:
I removed underscores.

But I found problem. DL4JTools need dependency to nd4j-backends. 
If I prepare report (for DL4JTools) showing content of MultiLayerNetwork and Layers will be another problem. This has no solution in nd4j-common.
Could be better create for universal tools stand-alone module in highest level (there is e.g. deeplearning4j, nd4j, ...).

agibsonccc:
@clavvis no it doesn't. nd4j-common is used by the whole project. Not sure where you're getting that from.

clavvis:
I need to add in pom.xml depenedency to nd4j-backends. Can I?

clavvis:
I only add dependency to nd4j-api and all works.

agibsonccc:
Great! Yeah, it shouldn't need a backend. A backend is only needed at test scope time.

PR #2231, opened 2017-10-30
setShape and setStride are deprecated

gagatust:
We can't reshape ndarray by using setShape()

```java
INDArray ar = Nd4j.create(new double[]{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12});
INDArray ar2 =  ar.reshape(new int[]{3,2,2}); //work
ar.setShape(new int[]{3,2,2}); //java.lang.IllegalStateException: Shape and stride must be the same length
```
https://github.com/deeplearning4j/nd4j/blob/700fe97d83d8f56ef09bb4246bcd679a272f8b64/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/indexing/NDArrayIndex.java#L133-L145

https://github.com/deeplearning4j/nd4j/blob/700fe97d83d8f56ef09bb4246bcd679a272f8b64/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java#L1045-L1055


raver119:
Looks good, thanks.

PR #2232, opened 2017-10-31
Add tags goals to tag goal

clavvis:

In pom.xml was one case of tag goal without tags goals.
In case of import in Eclipse it get error:
Cannot parse lifecycle mapping metadata for maven project MavenProject: org.nd4j:nd4j:0.9.2-SNAPSHOT @ C:\Java_ws_E2\dl4j\dl4j\nd4j\pom.xml Cause: Unrecognised tag: 'goal' (position: START_TAG seen ...</versionRange>\r\n        <goal>... @9:15)

## How was this patch tested?

Check it manually.


clavvis:
I did it bad. I had to do it better.

PR #2253, opened 2017-11-09
Add JavaCPP extensions for AVX512 to Nd4jCpuPresets

saudet:

At build time, when setting the "javacpp.extension" property to "-avx512", for example, `mvn -Djavacpp.extension=-avx512 ...`, it will link against the binaries built for AVX512 by libnd4j, and create artifacts ending with "-avx512", for example `nd4j-native-0.9.2-SNAPSHOT-linux-x86_64-avx512.jar` on Linux.

Then, when users wish to "enable" AVX512, they simply need to add the corresponding dependency, for example:
```xml
        <dependency>
            <groupId>org.nd4j</groupId>
            <artifactId>nd4j-native</artifactId>
            <version>${nd4j.version}</version>
            <classifier>linux-x86_64-avx512</classifier>
        </dependency>
```
And JavaCPP will try to load that one first.

## How was this patch tested?

Build still passes with and without "-Djavacpp.extension=-avx512", producing the correct binaries. Running some sample code, it correctly terminates with a SIGILL when trying to use the AVX512 artifact on my CPU (because it does not support AVX512) and runs fine without adding the new artifact.

Still need to add some code for detecting CPU features, but that will come later.


saudet:
FYI, @raver119 basically approved this in https://github.com/deeplearning4j/libnd4j/pull/554.

PR #2267, opened 2017-11-14
Exclude Junit from neoitertools, add the correct dependency where it should be.

### LINKED ISSUE #2262: Neoitertools dependency leaks transitive (and old) compile scope JUnit to ND4J
https://search.maven.org/#artifactdetails%7Cnet.ericaro%7Cneoitertools%7C1.0.0%7Cjar

```
<dependencies>
        <dependency>
            <groupId>junit</groupId>
            <artifactId>junit</artifactId>
            <version>4.8.2</version>
            <type>jar</type>
            <scope>compile</scope>
        </dependency>
    </dependencies>
```

huitseeker:
Fixes #2262


saudet:
We should still check in Nd4jCpu.java somewhere though, it gets put in the docs, etc.

PR #2275, opened 2017-11-17
Call overloaded initializeDevicesAndFunctions() to initialize special BLAS functions

saudet:

Switches the initializeDevicesAndFunctions() function call to the new overloaded one to pass functions pointers we're interested in.

## How was this patch tested?

Build passes, tests run OK.


raver119:
I'll add few tests later, once i attach these pointers to actual ops.

PR #2289, opened 2017-11-21
delete misleading  method

gagatust:
"Create" method does't create ndarray but only returns the argument.


raver119:
Does nd4j compiles fine with this change?

gagatust:
nd4j-api compiles. This method is not used.

PR #2317, opened 2017-11-28
refactoring: create autoProcessScalarCall

gagatust:


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java
```diff
@@ -4933,6 +4920,11 @@ protected void checkArrangeArray(int[] arr) {
         }
 
     }
+    
+    protected void autoProcessScalarCall() {
```
huitseeker: If you add this one as protected, rather than private, it becomes very tempting to ask you to add a unit test.

gagatust:
@huitseeker 
I wanted to add this one as private but I found a BaseSparseNDArrayCOO class that has the same statements.
https://github.com/deeplearning4j/nd4j/blob/b554e7fc41a09b82f44d4e11690639fd849bdb28/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseSparseNDArrayCOO.java#L233-L234

https://github.com/deeplearning4j/nd4j/blob/b554e7fc41a09b82f44d4e11690639fd849bdb28/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseSparseNDArrayCOO.java#L611-L612

BaseSparseNDArrayCOO has TODO – “BaseSparseNDArray should extend from BaseNDArray : remove the duplicate methods”.
https://github.com/deeplearning4j/nd4j/blob/b554e7fc41a09b82f44d4e11690639fd849bdb28/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseSparseNDArrayCOO.java#L29-L37
autoProcessScalarCall() could be used in BaseSparseNDArrayCOO. That is why it’s not private.
It can be package-private (no explicit modifier) if these classes will always be in one package.

agibsonccc:
Protected was fine. It's good for the base implementations. Thanks for the pull request @gagatust

PR #2351, opened 2017-12-05
Add optional class

AlexDBlack:
Based on the API of Java 8's Optional class. Required in DL4J here: https://github.com/deeplearning4j/deeplearning4j/pull/4370


saudet:
nd4j-common depends on Java 1.8:
https://github.com/deeplearning4j/nd4j/blob/master/nd4j-common/pom.xml#L14
Maybe we could just use that one?

AlexDBlack:
Huh, I didn't realize that.
But that opens another discussion - it means *everything* in the stack is by default now Java 8... nd4j-common -> nd4j-context -> nd4j-api -> all of DL4J. Which is probably not what we want.
I'll open an issue for that.

PR #2355, opened 2017-12-06
Generate protobuf dependencies at compilation

huitseeker:
This should let us avoid checking in 130K+ lines of tool-generated code, and should make the maintenance 
much easier, by reflecting the changes on the protocol definition files directly.

(and to preempt the question, *technically* the protobuf definition don't necessarily need to be there, but the git-fu necessary to link to them will be in another PR)


huitseeker:
@agibsonccc I address this point precisely in the commit message of the first commit. This change simply reflects updates in Onnx. The protobuf definitions imported here are from Onnx' master at https://github.com/onnx/onnx/commit/ecf87c3dddbc8362a826c84360e0797075e0da90 where `TensorTypeProto` does not exist.

agibsonccc:
@huitseeker No seriously, we should not be using TensorProto. We should be using TensorTypeProto. We need shape information. I'm not merging this till the OnnxGraphMapper is handled properly.

huitseeker:
@agibsonccc We are using and have shape information — otherwise the following sort of call would not compile:
```
    public int[] getShapeFromTensor(onnx.OnnxProto3.TypeProto.Tensor tensorProto) {
        val ret = new int[Math.max(2,tensorProto.getShape().getDimCount())];
        int dimCount = tensorProto.getShape().getDimCount();
```

https://github.com/huitseeker/nd4j/blob/413acddca63eb2f07d74f954d675205452e4ad1d/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/imports/graphmapper/onnx/OnnxGraphMapper.java#L328-L330

Once again, please have a cursory look at the renaming that occurs here: https://github.com/onnx/onnx/pull/215/files

huitseeker:
@agibsonccc Please read the above. I think you're missing a point. It's not that I choose to not use `TensortTypeProto`. It's that it's been renamed, and you're missing that it's the exact same object under a different name (differentiated from the `Tensor` you think about by the package name).

Issue #2352, opened 2017-12-05
nd4j-common is Java 8 therefore everything is Java 8


AlexDBlack:
https://github.com/deeplearning4j/nd4j/blob/master/nd4j-common/pom.xml#L14

Dependency chain:
nd4j-common -> nd4j-context -> nd4j-api

Unless (and until) we've decided to migrate to Java 8 for everything, this should stay as Java 7.

AlexDBlack:
Looking through Java 8 dependencies (via mvn site):

```
Java 8:
nd4j-api
nd4j-native-api		(graphviz-java-0.1.4, nd4j-api, nd4j-common)
nd4j-common  		(none - compile is java 8)
nd4j-context 		        (nd4j-common)
nd4j-buffer  		        (nd4j-common)
nd4j-perf	 		        (nd4j-common)
nd4j-base64  		        (nd4j-api, nd4j-common)
nd4j-jdbc	 		        (graphviz-java-0.1.4, nd4j-api, nd4j-common)
nd4j-serde			(graphviz-java-0.1.4, nd4j-api, nd4j-common)
nd4j-api			        (graphviz-java-0.1.4, nd4j-common)
nd4j-jdbc-mysql		(graphviz-java-0.1.4, nd4j-api, nd4j-common)
nd4j-cuda-platform	(graphviz-java-0.1.4, nd4j-api, nd4j-common)
nd4j-serde			(graphviz-java-0.1.4, aeron-all-1.04, nd4j-api, nd4j-common)
nd4j-gson			(graphviz-java-0.1.4, nd4j-api, nd4j-common)
nd4j-cuda			(graphviz-java-0.1.4, nd4j-api, nd4j-common, spring-core-5.0.2.RELEASE, spring-jcl-5.0.2.RELEASE)
nd4j-native			(graphviz-java-0.1.4, nd4j-api, nd4j-common, spring-core-5.0.2.RELEASE, spring-jcl-5.0.2.RELEASE)
nd4j-kafka			(graphviz-java-0.1.4, nd4j-api, nd4j-common)

Thinsg that should be Java 8, with multiple java 8 dependencies:
nd4j-parameter-server-rocksdb-storage
```

saudet:
I think a lot of that was updated when @agibsonccc merged samediff in nd4j-api...

AlexDBlack:
Yup, doesn't look too bad... really it's just the graphviz + those 2 spring dependencies that actually need java 8

PR #2356, opened 2017-12-06
Back to Java 7

AlexDBlack:
Fixes: https://github.com/deeplearning4j/nd4j/issues/2352

Tested locally: this builds, and I checked the impacted dependencies via mvn site, looks like everything that is supposed to be java 7 is back to java 7 (nd4j-native and nd4j-cuda still show java 8 in mvn site, but the dependencies are test scope).

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/pom.xml
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/pom.xml
@@ -39,8 +39,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>1.7</source>
+                    <target>1.7</target>
                 </configuration>
             </plugin>
             <plugin>
@@ -99,12 +99,6 @@
             <groupId>com.google.protobuf</groupId>
             <artifactId>protobuf-java</artifactId>
             <version>3.4.0</version>
-        </dependency>
-
-        <dependency>
-            <groupId>guru.nidi</groupId>
-            <artifactId>graphviz-java</artifactId>
-            <version>0.1.4</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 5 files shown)


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/layers/Linear.java
```diff
@@ -43,7 +44,7 @@ public Linear(int nIn,
         super(null,
                 getParams(nIn,nOut,weightInitScheme,biasWeightInitScheme),
                 new INDArray[]{},
```
agibsonccc: this is java *6* we should be fine without this.

nd4j-backends/nd4j-api-parent/nd4j-api/pom.xml
```diff
@@ -39,8 +39,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>1.7</source>
```
agibsonccc: yes this makes sense. I want java 8 in the tests now if we can. As for samediff there might have been some modifications there.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/layers/Linear.java
AlexDBlack: I thought so too - but Intellij was complaining about it, so I changed it. I just checked with maven now - it's unhappy there too, without it: https://gist.github.com/AlexDBlack/1f76e216bc26d3946d4bf2061a045e2f

saudet:
@AlexDBlack @agibsonccc
Should we get this done before the coming release?
Or is it OK to have Java 8 as requirement?

agibsonccc:
Yeah..We shouldn't be all in on java 8 *yet*

PR #2358, opened 2017-12-06
R119 notes

raver119:
***WIP; DO NOT MERGE;***

Adding missing wrappers for new operations available in libnd4j


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/shape/MergeSum.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        if (idx < 0)
+            throw new ND4JIllegalStateException("Can't find dimension for concatenatiion");
+
+        // if that's convolution graph, we should swap dimensions
+        if (concatDimension == 3)
+            concatDimension = 1;
+
+        this.concatDimension = concatDimension;
```
agibsonccc: missing an addIArgument call here.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/transforms/Erf.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public String opName() {
+        return "erf";
+    }
+
+    @Override
+    public String onnxName() {
+        return "Erf";
```
agibsonccc: Look here for onnx equivalents:  https://github.com/onnx/onnx/blob/master/docs/Operators.md If they don't exist then throw an exception.

PR #2373, opened 2017-12-13
Add tests for MathUtils & update to DL4J status

huitseeker:

Ports the test from:
https://github.com/deeplearning4j/deeplearning4j/pull/4242

After updating this copy of Mathutils

### DIFF
--- a/nd4j-common/src/main/java/org/nd4j/linalg/util/MathUtils.java
+++ b/nd4j-common/src/main/java/org/nd4j/linalg/util/MathUtils.java
@@ -14,33 +14,34 @@
  *  *    See the License for the specific language governing permissions and
  *  *    limitations under the License.
  *
- *
  */
 
 package org.nd4j.linalg.util;
 
-
+import org.apache.commons.math3.linear.CholeskyDecomposition;
+import org.apache.commons.math3.linear.NonSquareMatrixException;
+import org.apache.commons.math3.linear.RealMatrix;
 import org.apache.commons.math3.random.RandomGenerator;
 import org.apache.commons.math3.util.FastMath;
+import org.nd4j.linalg.primitives.Counter;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
+import java.util.Set;
 
 
 /**
- * This is a math jcuda.utils class.
+ * This is a math utils class.
  *
  * @author Adam Gibson
+ *
  */
 public class MathUtils {
 
-
-
-    /**
-     * The natural logarithm of 2.
-     */
+    /** The natural logarithm of 2. */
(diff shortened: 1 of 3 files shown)


nd4j-common/src/main/java/org/nd4j/linalg/util/MathUtils.java
```diff
@@ -1168,12 +1224,11 @@ public static int distanceFinderZValue(double[] vector) {
...
 
     /**
-     * This returns the euclidean distance of two vectors
+     * This returns the distance of two vectors
      * sum(i=1,n)   (q_i - p_i)^2
-     *
      * @param p the first vector
      * @param q the second vector
-     * @return the euclidean distance between two vectors
+     * @return the distance between two vectors
```
maxpumperla: Why would we want to be less specific intentionally?

nd4j-common/src/main/java/org/nd4j/linalg/util/MathUtils.java
```diff
@@ -239,34 +269,38 @@ public static double vectorLength(double[] vector) {
...
-     *
-     * @param totalDocs                       the total documents for the data applyTransformToDestination
+     * @param totalDocs the total documents for the data applyTransformToDestination
      * @param numTimesWordAppearedInADocument the number of times the word occurred in a document
      * @return log(10) (totalDocs/numTImesWordAppearedInADocument)
      */
     public static double idf(double totalDocs, double numTimesWordAppearedInADocument) {
-        return totalDocs > 0 ? Math.log10(totalDocs / numTimesWordAppearedInADocument) : 0;
+        //return totalDocs > 0 ? Math.log10(totalDocs/numTimesWordAppearedInADocument) : 0;
+        if (totalDocs == 0)
```
maxpumperla: is ternary op less efficient? not sure of the value here.

Can we remove that line instead of commenting out?

also, why not directly `return Math.log10(totalDocs / numTimesWordAppearedInADocument);`?

nd4j-common/src/main/java/org/nd4j/linalg/util/MathUtils.java
```diff
@@ -239,34 +269,38 @@ public static double vectorLength(double[] vector) {
...
     /**
      * Term frequency: 1+ log10(count)
-     *
      * @param count the count of a word or character in a given string or document
      * @return 1+ log(10) count
      */
-    public static double tf(int count) {
-        return count > 0 ? 1 + Math.log10(count) : 0;
+    public static double tf(int count, int documentLength) {
+        //return count > 0 ? 1 + Math.log10(count) : 0
```
maxpumperla: same as above

PR #2421, opened 2017-12-21
Add .travis.yml and build scripts for Linux and Mac OS X

saudet:

Let Travis CI make builds and deploy SNAPSHOT artifacts to Sonatype OSS automatically for CPU and CUDA on Linux and Mac OS X.

## How was this patch tested?

Builds for Linux and Mac OS X pass on Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,14 @@ services:
 matrix:
   include:
     - os: linux
+      env: OS=android-arm SCALA=2.10
+      install: true
+      script: bash ./ci/build-android.sh
+    - os: linux
+      env: OS=android-x86 SCALA=2.11
+      install: true
+      script: bash ./ci/build-android.sh
+    - os: linux
       env: OS=linux-x86_64 CUDA=8.0 CUDNN=6 SCALA=2.10
       install: true
       script: bash ./ci/build-linux-x86_64.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+dist: trusty
+sudo: required
+cache:
+  directories:
+    - $HOME/.m2
+language: java
+services:
+  - docker
+matrix:
+  include:
+    - os: linux
+      env: OS=linux-x86_64 CUDA=8.0 CUDNN=6 SCALA=2.10
+      install: true
+      script: bash ./ci/build-linux-x86_64.sh
+    - os: linux
+      env: OS=linux-x86_64 CUDA=9.0 CUDNN=7 SCALA=2.11
+      install: true
+      script: bash ./ci/build-linux-x86_64.sh
+    - os: osx
+      osx_image: xcode7.3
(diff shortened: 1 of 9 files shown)


ci/build-macosx-x86_64.sh
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+brew update > /dev/null
+brew install gcc || true
+brew link --overwrite gcc
+
+if [[ $CUDA == "8.0" ]]; then
+    curl -L https://developer.nvidia.com/compute/cuda/8.0/Prod2/local_installers/cuda_8.0.61_mac-dmg -o $HOME/cuda.dmg
+else
+    curl -L https://developer.nvidia.com/compute/cuda/9.0/Prod/local_installers/cuda_9.0.176_mac-dmg -o $HOME/cuda.dmg
```
agibsonccc: We might want to host these ourselves.

.travis.yml
```diff
@@ -0,0 +1,29 @@
+dist: trusty
```
agibsonccc: Isn't fedora required? I noticed that the build down below uses RPMs.

ci/build-linux-x86_64.sh
```diff
@@ -0,0 +1,34 @@
...
+    MAVEN_PHASE="install"
+fi
+
+if ! git -C $TRAVIS_BUILD_DIR/.. clone https://github.com/deeplearning4j/libnd4j/ --depth=50 --branch=$TRAVIS_BRANCH; then
+     git -C $TRAVIS_BUILD_DIR/.. clone https://github.com/deeplearning4j/libnd4j/ --depth=50
+fi
+
+docker run -ti -e SONATYPE_USERNAME -e SONATYPE_PASSWORD -v $HOME/.m2:/root/.m2 -v $TRAVIS_BUILD_DIR/..:/build \
+    nvidia/cuda:$CUDA-cudnn$CUDNN-devel-centos6 /bin/bash -evxc "\
+        yum -y install centos-release-scl-rh epel-release; \
```
agibsonccc: Flagging for above, it seems like we're using ubuntu but we're using RHEL commands down here.

.travis.yml
saudet: Travis only supports Trusty, that's why we need to use Docker.

ci/build-linux-x86_64.sh
saudet: Yes, those commanes are running it Docker.

ci/build-macosx-x86_64.sh
saudet: We'll see, I guess? Going through Microsoft, Amazon, or NVIDIA, not sure there is a big difference...

saudet:
I got builds working for Android as well!

@agibsonccc So, anything that we need to change before we can merge this?

PR #2473, opened 2018-01-05
Add JavaCPP extension for AVX2 to Nd4jCpuPresets and CI builds

saudet:

Add necessary modifications to support AVX2 and to build on AppVeyor and Travis CI.

## How was this patch tested?

Builds and runs fine locally, but wait until we get the green light from the CI services before merging.


saudet:
Yes, that's the idea. Starting with AVX2 only with CI for now, but AVX512 should be working as well.

PR #2511, opened 2018-01-11
adds short serialisation section.

RobAltena:
adds short serialisation section.


RobAltena:
Well that went horribly wrong.

PR #2512, opened 2018-01-11
adds short serialisation section.

RobAltena:
adds short serialisation section.

### DIFF
--- a/userguide.md
+++ b/userguide.md
@@ -35,6 +35,7 @@ This user guide is designed to explain (and provide examples for) the main funct
   * Permute
   * sortRows/sortColumns
   * Directly accessing BLAS operations
+* <a href="#serialization">Serialization</a>
 * <a href="#quickref">Quick Reference: A Summary Overview of ND4J Methods</a>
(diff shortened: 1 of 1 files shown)


AlexDBlack:
Thanks!
So one possible improvement here: Nd4j.read(InputStream) does the wrapping with the DataInputStream as required.
Also maybe note that the streams should be manually closed (or use try-with-resources etc).

agibsonccc:
May also want to mention:
https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/serde/binary/BinarySerde.java

as well as:
https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4j.java#L2265

There's various serialization modules as well:
https://github.com/deeplearning4j/nd4j/tree/master/nd4j-serde

AlexDBlack:
Thanks!

PR #2555, opened 2018-01-19
Update to JavaCPP 1.4 and add support for CUDA 9.1

saudet:

Update to the latest release versions of JavaCPP and CUDA.

## How was this patch tested?

Unit tests pass, but let's wait after the CI server for the build scripts...


saudet:
@raver119 Looks good? Windows builds are failing right now, but that seems to be a problem in libnd4j...

raver119:
I've fixed cuda on windows few hours ago, @AlexDBlack confirmed that

saudet:
@raver119 Ok, cool, so please merge this if it looks good! Thanks

PR #2564, opened 2018-01-23
Add libnd4j-assembly Maven profile to unpack from dependencies

saudet:

Complement of https://github.com/deeplearning4j/libnd4j/pull/702. When executing a build with say `mvn clean install -Plibnd4j-assembly -Djavacpp.extension=avx2`, it will not require libnd4j to be available anywhere. It will pick up assemblies from Maven dependencies, unpack in the private `target` directory, and use its binaries automatically.

@sshepel BTW, if I were to refactor nd4j-native and nd4-cuda, I think I'd try to use classifiers for different versions of CUDA. Having to change the version with a script isn't necessary like it is with Scala because in case of CUDA, the compiled Java classes are the same. So, I think it's a good idea to start doing it that way for libnd4j at least. What do you think?

## How was this patch tested?

Build manually tested for nd4j-native and nd4j-cuda.

### DIFF
--- a/nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom.xml
+++ b/nd4j-backends/nd4j-backend-impls/nd4j-cuda/pom.xml
@@ -271,6 +271,73 @@
                 </plugins>
             </build>
         </profile>
+
+        <profile>
+            <id>libnd4j-assembly</id>
+            <activation>
+                <property>
+                    <name>libnd4j-assembly</name>
+                </property>
+            </activation>
+            <properties>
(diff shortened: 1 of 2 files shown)


raver119:
@saudet how sbt will play with classifiers in such scenario?

saudet:
It's only used at build time for ND4J, not for any of other project using it as a dependency.

saudet:
Ah, yes it should work fine, but then again it might not be a good idea because the dependencies change with the version of CUDA (presets for CUDA 8.0 vs 9.0). We should probably keep doing as with Scala :)

PR #2594, opened 2018-02-02
Fix for obtaining own IP by parameter server client

tomkozicki:

VoidParameterServer now doesn't assume SPARK_PUBLIC_DNS as it's IP first.

## How was this patch tested?

Ran MNIST parameter server example after building this and analogous deeplearning4j change with expected result


tomkozicki:
Should be considered along with:
https://github.com/deeplearning4j/deeplearning4j/pull/4599/commits/ed7fe2122e60f60d3b926b4d02600cf8f4822fe5

PR #2702, opened 2018-03-05
Return *mutable* lists from doDiff

treo:
`Collections.singletonList` returns an immutable list, yet `DifferentialFunction.diff` expects that the list is mutable. If it isn't, it blows up when trying to run `SameDiff.execBackwards`.

`doDiff` returns mutable lists in other contexts as well, so this shouldn't violate any implicit assumptions.


AlexDBlack:
> DifferentialFunction.diff expects that the list is mutable

Perhaps we should add a check to DifferentialFunction class for this? (i.e., a try/catch with useful exception)... it's way too easy to return a non-mutable list in practice then have to spend time debugging it. TBH I might even consider allowing users to return anything, and handle non-mutability in execBackwards - that would be easier for users and less error prone...

Also, Arrays.asList is mutable in that you can replace elements via set(int, T) but not add elements via add(T)... not sure if we need that functionality?

treo:
https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/functions/DifferentialFunction.java#L516
```
 public List<SDVariable> diff(List<SDVariable> i_v1) {
        List<SDVariable> vals = doDiff(i_v1);
        if(vals == null){
            throw new IllegalStateException("Error executing diff operation: doDiff returned null for op: " + this.opName());
        }

        val outputVars = args();
        for(int i = 0; i < vals.size(); i++) {
            SDVariable var = outputVars[i];
            SDVariable grad = var.getGradient();
            if(grad != null) {
                SDVariable gradVar =  f().add(grad, vals.get(i));
                vals.set(i, gradVar);
                sameDiff.setGradientForVariableName(var.getVarName(), gradVar);
            } else {
                SDVariable gradVar = vals.get(i);
                sameDiff.updateVariableNameAndReference(gradVar,var.getVarName() + "-grad");
                sameDiff.setGradientForVariableName(var.getVarName(), gradVar);
                sameDiff.setForwardVariableForVarName(gradVar.getVarName(),var);

            }
        }

        return vals;
    }
```

It uses only `.set` in there, and only in a special case. I've used Arrays.asList as it was used in other places where more than a single element was returned. 

Handling the case in there might also be viable, I just wanted to skip a copy :)

AlexDBlack:
> It uses only .set in there, and only in a special case. I've used Arrays.asList as it was used in other places where more than a single element was returned.

Yeah, ok. Sounds reasonable.

At the very least, I still want a try/catch (with useful exception) on the .set() call though, to catch the immutable case. People should be able to implement their own ops without having to understand and debug issues like that.

treo:
@AlexDBlack something like that?

AlexDBlack:
@treo yep, great 👍

PR #2708, opened 2018-03-06
Fix expandDims

treo:
The original implementation only worked for a single case: dimensions = 1

Also I don't see a good reason to use Indexing, if a simple reshape should work just fine.


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/factory/Nd4j.java
```diff
@@ -401,11 +399,11 @@ public static INDArray prepend(INDArray arr, int padAmount, double val, int axis
...
             dimension += input.rank();
-        INDArrayIndex[] indexes = new INDArrayIndex[input.rank()];
+        int[] shape = input.shape();
+        int[] indexes = new int[input.rank() + 1];
         for (int i = 0; i < indexes.length; i++)
-            indexes[i] = NDArrayIndex.all();
-        indexes[dimension] = NDArrayIndex.newAxis();
-        return input.get(indexes);
+            indexes[i] = i < dimension ? shape[i] : i == dimension ? 1 : shape[i - 1];
+        return input.reshape(indexes);
```
AlexDBlack: Hm... two thoughts here:
(a) reshape isn't always a no-copy op. Not sure, but we might end up returning a copy in some cases with this approach?
(b) if we do use reshape, I'd be more comfortable with a .reshape(input.ordering(), indexes)

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/linalg/factory/Nd4jTest.java
```diff
@@ -122,5 +123,13 @@ public void testVar2() {
         assertEquals(Nd4j.create(new double[] {2.25, 2.25, 2.25}), var);
     }
 
+    @Test
+    public void testExpandDims(){
```
AlexDBlack: Test is a good start - but we really need to test the 'weird stride' cases IMO...
Use this + loop over all test cases: https://github.com/deeplearning4j/nd4j/blob/88f62abd5b43489942f4394bf90f901de5b09a17/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/checkutil/NDArrayCreationUtil.java#L25-L30

Furthermore, I'd prefer if we checked the acutal values, *not* just the shapes here.
And also test whether the result is still a view in all cases.

(Yes, this is way more detailed testing than we originally had, but it's necessary IMO)

treo:
I've tried to make it Work with NDArrayIndex, but even after fixing the wrong position of `newAxis` it still returned the wrong shape. Maybe there is a bug in there too?

This updated version now uses more thorough tests and preserves the input ordering.

AlexDBlack:
Looks good, assuming tests pass... so my only remaining question is whether we want to enforce the expanded one always being a view - and if so, we should also test for that. I'm inclined to do that, but could perhaps be persuaded otherwise
(one simple way to test: modify the original, check the expanded one also changes)

treo:
According to `.isView()` the result isn't a view, yet when I run the following, it doesn't fail.
```
testMatrix.assign(Nd4j.rand(shape));
assertEquals(message, testMatrix.ravel(), expanded.ravel());
```

I'm confused a bit by that.

AlexDBlack:
@treo it's possible it's not *always* a view... some of the arrays returned by NDArrayCreationUtil aren't views (but most are). If that's not simply what's going on here - maybe open an issue.
I'm happy with the assign + assert approach though.

treo:
@AlexDBlack I meant that `expanded.isView()` returns false, yet, assign to `testMatrix` changes it.

AlexDBlack:
@treo yes, that was my interpretation and what I was referring to. Specifically, you could legitimately imagine that Nd4j.create(10,10).transpose().isView() could return false (not sure if it actually does) - it's contiguous in buffer and is offset 0...

treo:
@AlexDBlack Done.

Issue #2632, opened 2018-02-13
Windows - CUDA initialization blocked?


AlexDBlack:
Not sure what's up here, seems to just get stuck on initializeDevicesAndFunctions...
![image](https://user-images.githubusercontent.com/2360237/36147242-91f2930a-110b-11e8-9058-9582a4f65d34.png)

CUDA test profile under windows 10. Attempting to run simple DL4J unit test. IntelliJ invalidate and restart doesn't help...

libnd4j: 4527181f7625b2312e6d631c3e48e336bf0e9262
nd4j: 90b1ef0fe7d2cf8f3081bf896e8822c9ab0faf66

AlexDBlack:
Closing this for now - I haven't been able to reproduce this problem since... will re-open if it occurs again.

AlexDBlack:
This has resurfaced for me... I'm seeing it again after building today.
Same exact method. Looks like Java is maxing out one single thread here with initializeDevicesAndFunctions()
![image](https://user-images.githubusercontent.com/2360237/36879847-c99ca082-1e19-11e8-96a1-9d80bc1728a1.png)

saudet:
This happens even when building locally for your compute capability?

AlexDBlack:
yeah, this is a fresh build today on Windows, running in IntelliJ (doesn't appear to be using sonatype snapshots as far as I can tell)

raver119:
Hm. But i've used windows yesterday to test your issue - no problems at all, CUDA works for me as expected.

raver119:
compilation params were the same as always: `-a native -c cuda -cc 61`

AlexDBlack:
Right, and I've been rebuilding ND4J and DL4J every day (no ```-a native``` though) - and I'm only seeing it on occasion (not sure what the trigger is). It's been fine for me, for the last week or so (but obviously saw issues before that).

AlexDBlack:
And after rebuilding today - it seems to be fine again... I'm inclined to keep this open until I know what's going on here and why.

AlexDBlack:
It resurfaced today after building again in exactly the same way as before. I let it run for 2 hours without seeing any evidence of progress. Note that I've hit this on both CUDA 9.0 and 9.1 now. I'll rebuild and use CUDA 8.0 for a while, and see if it occurs there.

wmeddie:
I also run into this 100% of the time on a V100.  Both in the SKIL version on 0.9.2 and the latest snapshots.

[...]

PR #2716, opened 2018-03-08
Log about initializeDevicesAndFunctions() potentially taking a while

### LINKED ISSUE #2715: CUDA runtime (ptx) compliation - detect when occuring, log message
With current snapshots, we apparently aren't building for all architectures. For other architectures, a compilation step is required. This can take a considerable amount of time (10+ minutes?).

Currently, this provides zero feedback to the user when this is occuring - it just looks like ND4J has frozen during initialization, which provides a very poor user experience.

saudet:
Fixes #2715 

Add call to log.info() before call to initializeDevicesAndFunctions()

## How was this patch tested?

Tested manually: Message shows with nd4j-cuda backend, doesn't show with nd4j-native one.

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-native-api/src/main/java/org/nd4j/nativeblas/NativeOpsHolder.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-native-api/src/main/java/org/nd4j/nativeblas/NativeOpsHolder.java
@@ -29,6 +29,9 @@ public class NativeOpsHolder {
             Class<? extends NativeOps> nativeOpsClazz = Class.forName(name).asSubclass(NativeOps.class);
             deviceNativeOps = nativeOpsClazz.newInstance();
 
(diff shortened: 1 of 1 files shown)


saudet:
Let's just add a note about that in https://deeplearning4j.org/snapshots then?

AlexDBlack:
Well, again, it's bad UX... we already know people don't read docs - and even those that do won't necessarily remember. I'd rather we make it obvious/explicit/automatic as possible. I'm happy to implement the timer-based approach I mentioned, if you want.

saudet:
But I don't like these ugly hacks :) They create additional bugs we don't need...

Well, since CUDA 9 can't reliably JIT compile libnd4j anymore (issue #2632) maybe this is a moot point?

AlexDBlack:
@saudet you might be right about that... until we have JIT compilation occurring reliably - or (even better :)) we don't have any cases where JIT is required - warning won't do much to help.

PR #2721, opened 2018-03-09
Fix BaseLevel3.gemm order arg (get from GemmParams in case of duping)

AlexDBlack:
***DO NOT MERGE***
Attempts to fix: https://github.com/deeplearning4j/deeplearning4j/issues/4776#issuecomment-371745349

Order arg should be obtained from GemmParams.
Note however this is *not* sufficient to fix the afore-mentioned issue.


raver119:
Is this still viable?

AlexDBlack:
I completely forgot about this. The alpha/beta stuff there (instead of 0s and 1s) should probably be implemented/fixed. Maybe I'll close this and do a separate PR for that.

PR #2722, opened 2018-03-09
Fix issue where configuration ExecutionModel was set to SEQUENTIAL re…

enragedginger:
…gardless of the value passed in.

### DIFF
--- a/nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/jita/conf/Configuration.java
+++ b/nd4j-backends/nd4j-backend-impls/nd4j-cuda/src/main/java/org/nd4j/jita/conf/Configuration.java
@@ -607,7 +607,7 @@ public class Configuration implements Serializable {
      * @return
      */
     public Configuration setExecutionModel(@NonNull ExecutionModel executionModel) {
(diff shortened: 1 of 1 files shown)


AlexDBlack:
Thanks!

raver119:
That's not a bug, and this PR should be reverted. Sequential execution is the only available option atm. Other approaches were deprecated and should be removed.

enragedginger:
If indeed that's the case, then I don't think simply reverting this PR is the best course of action. Providing a "settter" method which sets an object's property to any value other than what was passed in is a very bad coding practice as it leads to surprising behavior.

We should probably do one of the following things instead:
1. Actual implement the other possible values.
2. Remove the method entirely. There can be only one possible value, so why bother providing such a useless method? If we go with this route, then the comments / docs should be updated to reflect why this path was chosen.

Given that you're infinitely more familiar with the project, what are your thoughts on this?

raver119:
2. There were other ways to handle execution. But over time Sequential became hybrid of all other methods - code from other executors/controllers migrated into it.  So ye, we should just remove that stuff eventually.

AlexDBlack:
See: https://github.com/deeplearning4j/nd4j/pull/2760

PR #2724, opened 2018-03-10
Implement loss function for multi label classification tasks

treo:
This implements the loss function from this paper:

Min-Ling Zhang and Zhi-Hua Zhou, "Multilabel Neural Networks with Applications to Functional Genomics and Text Categorization," in IEEE Transactions on Knowledge and Data Engineering, vol. 18, no. 10, pp. 1338-1351, Oct. 2006.

It is useful as a loss function for multi label classification tasks.

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMultiLabel.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMultiLabel.java
@@ -20,6 +20,7 @@ import org.tensorflow.framework.AttrValue;
 import org.tensorflow.framework.GraphDef;
 import org.tensorflow.framework.NodeDef;
 
+import javax.naming.OperationNotSupportedException;
 import java.util.List;
 import java.util.Map;
 
@@ -181,17 +182,17 @@ public class LossMultiLabel extends DifferentialFunction implements ILossFunctio
 
     @Override
     public SDVariable[] outputVariables() {
-        return new SDVariable[0];
+        throw new UnsupportedOperationException();
     }
 
     @Override
     public SDVariable[] outputVariables(String baseName) {
-        return new SDVariable[0];
+        throw new UnsupportedOperationException();
     }
 
     @Override
     public List<SDVariable> doDiff(List<SDVariable> f1) {
-        return null;
+        throw new UnsupportedOperationException();
     }
 
 
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMultiLabel.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMultiLabel.java
@@ -26,21 +26,17 @@ import java.util.Map;
 /**
  * Multi-Label-Loss Function, maybe more commonly known as BPMLL
  * <p>
- * <p>
  * This Loss function requires that the Labels are given as a multi-hot encoded vector. It doesn't require any special
  * Activation method, i.e. the network output doesn't have to be in any specific range.
- * <p>
  * <p>
  * The loss is calculated based on the classification difference on labels that the examples has, and those that it
  * doesn't have. Assume that each example has a set of labels, these labels are the positive set, the labels that do not
  * belong to the example are in the negative set. This loss function trains the network to produce a higher value for
  * labels that are in the positive set than those that are in the negative set.
  * <p>
- * <p>
  * For a more detailed explanation and the actual formulas, read the original paper by Zhang and Zhou. The
  * implementation on scoreArray is based on equation 3, while computeGradient is based on equation 11. The main
  * difference being that -(c_k - c_l) = (c_l - c_k) was used to simplify the calculations.
- * <p>
  * <p>
  * Min-Ling Zhang and Zhi-Hua Zhou, "Multilabel Neural Networks with Applications to Functional Genomics and Text
  * Categorization," in IEEE Transactions on Knowledge and Data Engineering, vol. 18, no. 10, pp. 1338-1351, Oct. 2006.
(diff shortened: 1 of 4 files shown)


saudet:
@sshepel Did you set some commands to force Jenkins to rebuild the branch without having to push new commits?

sshepel:
> @sshepel Did you set some commands to force Jenkins to rebuild the branch without having to push new commits?
There is a button in Jenkins UI to do that, I have already provided @treo with all required info, no additional forcing logic where added.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMultiLabel.java
```diff
@@ -0,0 +1,227 @@
...
+            final INDArray pairwiseSub = Transforms.exp(operandA.sub(operandB));
+
+            final INDArray selection = locPositive.transpose().mmul(locNegative);
+
+            final INDArray classificationDifferences = pairwiseSub.muli(selection).divi(locNormFactor);
+
+            if (scoreOutput != null) {
+                if (mask != null) {
+                    final INDArray perLabel = classificationDifferences.sum(0);
+                    LossUtil.applyMask(perLabel, mask);
```
AlexDBlack: This line looks incorrect at first glance... seems like you are applying the full (all examples) mask on a single row?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMultiLabel.java
```diff
@@ -0,0 +1,227 @@
...
+        }
+        final INDArray postOutput = activationFn.getActivation(preOutput.dup(), true);
+
+        final INDArray positive = labels;
+        final INDArray negative = labels.eq(0.0);
+        final INDArray normFactor = negative.sum(1).muli(positive.sum(1));
+
+
+        int examples = positive.size(0);
+        for (int i = 0; i < examples; i++) {
```
AlexDBlack: I haven't checked the math - but I assume there's no way to do this without the loop?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMultiLabel.java
```diff
@@ -0,0 +1,227 @@
...
+            final INDArray locCfn = postOutput.getRow(i);
+            final int[] shape = locCfn.shape();
+
+            final INDArray locPositive = positive.getRow(i);
+            final INDArray locNegative = negative.getRow(i);
+            final INDArray locNormFactor = normFactor.getScalar(i);
```
AlexDBlack: Why not getDouble? Scalar in INDArray seems to provide (minor) overhead with no benefit?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMultiLabel.java
```diff
@@ -0,0 +1,227 @@
...
+                    final INDArray perLabel = classificationDifferences.sum(0);
+                    LossUtil.applyMask(perLabel, mask);
+                    perLabel.sum(scoreOutput.getRow(i), 0);
+                } else {
+                    classificationDifferences.sum(scoreOutput.getRow(i), 0, 1);
+                }
+            }
+
+            if (gradientOutput != null) {
+                gradientOutput.getRow(i).assign(classificationDifferences.sum(0).add(classificationDifferences.sum(1).transposei().negi()));
```
AlexDBlack: Minor optimization: sum(0).add*i*

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMultiLabel.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    @Override
+    public SDVariable[] outputVariables(String baseName) {
+        return new SDVariable[0];
+    }
+
+    @Override
+    public List<SDVariable> doDiff(List<SDVariable> f1) {
+        return null;
```
AlexDBlack: I'd rather an UnsupportedOperationException

treo: I've copied that over from the other loss functions. Not quite sure what exactly is expected here, so I've left it the way that it is with the others.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMultiLabel.java
treo: I was trying to keep the data movement off-heap->heap->off-heap to a minimum. But now that you've asked I actually looked at what getScalar does, and seen that it does just call getDouble internally... So I'll change that.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMultiLabel.java
treo: I've failed to find a way that works without the loop. Every method that I could come up with, that skips the loop also calculates A LOT of unnecessary results. The speed hit isn't as bad as it may look like. It is at about 85% to 90% of BinaryXENT (e.g. 130 batches per second vs 150 batches per second).

If you have an idea how it could be done any faster, or any more vectorized, I'd love to hear it :)

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/lossfunctions/impl/LossMultiLabel.java
treo: You've got a point here, I've been moving around the masking stuff to find a proper place for it, and missed calling getRow here.

PR #2726, opened 2018-03-10
Add concat and tile ops

farizrahman4u:


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/samediff/SameDiff.java
```diff
@@ -2502,6 +2510,22 @@ public SDVariable rollAxis(SDVariable x, int axis) {
...
+    public SDVariable concat(int dimension, SDVariable... inputs) {
+        return concat(null, dimension, inputs);
+    }
+
+    /**
+     * @param iX
+     * @param repeat
+     * @return
+     */
+    public SDVariable tile(SDVariable iX, int[] repeat) { return tile(iX, repeat);}
```
treo: Looks a lot like `return tile(null, iX, repeat)` was meant and this would create an infinite recursion / stack overflow instead.

nd4j-backends/nd4j-tests/src/test/java/org/nd4j/autodiff/samediff/SameDiffTests.java
```diff
@@ -276,10 +276,37 @@ public void testTranspose() {
...
 
+    @Test
+    public void testTile() {
+        SameDiff sameDiff = SameDiff.create();
+        INDArray arr = Transforms.sigmoid(Nd4j.linspace(1, 4, 4));
+        SDVariable x = sameDiff.var("x", arr);
+        SDVariable result = sameDiff.tile(x, new int[]{2, 2});
+        assertArrayEquals(new int[]{2, 8}, result.getShape());
+
+    }
```
treo: Can you add some tests to verify that not only the shape, but also the contents are as expected?

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/functions/DifferentialFunctionFactory.java
```diff
@@ -643,6 +643,9 @@ public SDVariable reverse(SDVariable x, int... dimensions){
         return new Reverse(sameDiff(),x, dimensions).outputVariables()[0];
     }
 
+    public SDVariable reverse_sequence(SDVariable x) {
+        return new Reverse(sameDiff(), x, 1).outputVariables()[0];
```
maxpumperla: you need a new op `ReverseSequence` that wraps `reverse_sequence` from libnd4j. also, make sure to write a test for it.

PR #2735, opened 2018-03-14
toString and readTxt/writeTxt related fixes

eraly:
* Fixes Nd4j.readTxt/writeTxt to work with true scalars and vectors. 
* Fixes bug associated with col vectors
* Deprecates jsonReader/jsonWriter which had an @Ignore on the tests so was completely not supported previously.
* Added in a test utility to hit corner cases with shapes (this might be present somewhere else?)
* Improves formatting on toString to look more like numpy: 
  - col vectors will print like column vectors are expected too
  - when length of an array is longer than a 1000 elements it will get summarized to only print first and last three slice/elements in every dimension
  - will switch to scientific notation when max abs value > 1000 or min abs value < 0.0001 or the ratio of the two is > 1000
  - default print is now four decimal points
Also our printing was completely wonky for shapes like [1,2,1] or [1,1,2] (not just because we didn't print col vectors properly) with brackets and all that now matches numpy. [before](https://gist.github.com/eraly/c48d694757df93a85dbf8c08e0e9452c) vs [after](https://gist.github.com/eraly/4ab86b002887797aacd312ed562bbfbe) print examples that show this.

Other print examples[ here](https://gist.github.com/eraly/ce103c745b879cd521528026a18492bd)

Issue: We rely heavily on slice for the toString conversion and there is a bug that only affects f order with some cases of shape where the size(rank-1) == 1 and till this is fixed I internally do a dup to c order only for this slice..Same for tads which we use in readTxt


eraly:
And anyone who will have an opinion on printing these arrays please review.

AlexDBlack:
Thanks for looking at this, it's something we definitely need... 
Here's a simple test on this branch:

```
    @Test
    public void test(){
        INDArray arr = Nd4j.create(new double[][]{
                {-1,0,1,0},
                {-0.1, 0.1, -10, 10},
                {-1e-2, 1e-2, -1e2, 1e2},
                {-1e-3, 1e-3, -1e3, 1e3},
                {-1e-4, 1e-4, -1e4, 1e4},
                {-1e-8, 1e-8, -1e8, 1e8},
                {-1e-50, 1e-50, -1e50, 1e50},
        });

        System.out.println(arr);
    }
```

```
[[  -1.0000E0,   0.0000E0,   1.0000E0,   0.0000E0], 
 [ -1.0000E-1,  1.0000E-1,  -1.0000E1,   1.0000E1], 
 [ -1.0000E-2,  1.0000E-2,  -1.0000E2,   1.0000E2], 
 [ -1.0000E-3,  1.0000E-3,  -1.0000E3,   1.0000E3], 
 [ -1.0000E-4,  1.0000E-4,  -1.0000E4,   1.0000E4], 
 [ -1.0000E-8,  1.0000E-8,  -1.0000E8,   1.0000E8], 
 [-1.0000E-50, 1.0000E-50, -1.0000E50,  1.0000E50]]
```

That's definitely a step up on what we currently have... however I wonder if we can improve it further?
Notably, the E0 suffix actually makes it harder to read IMO for those small integer values and I'd be tempted to not zero pad.
Also being pedantic, I'd use lower-case ```e``` - shorter character than digits makes it easier to read IMO

So 0.0 would print as 0.0, not 0.0000E0 or 0.0000e0, and 10.0 would print as either 1.0e1 or 10.0 (I'm open to either, with slight preference for switching to scientific on a *per number* basis instead of a per-array basis)

eraly:
Thanks for the review Alex. I was following along numpy's convention to switch to scientific notation for the whole array and totally agree that we can improve on that :)
Will be sure to get this into our next release.

eraly:
```
[[       -1e0,        0e0,     1.0000,        0e0], 
 [      -1e-1,     0.1000,       -1e1,    10.0000], 
 [      -1e-2,     0.0100,       -1e2,   100.0000], 
 [      -1e-3,     0.0010,       -1e3,  1000.0000], 
 [      -1e-4,       1e-4,       -1e4, 10000.0000], 
 [      -1e-8,       1e-8,       -1e8,        1e8], 
 [       -0e0,        0e0,         -∞,          ∞]]
```

eraly:
When switching to scientific notation (or if the value is zero) I don't zero pad but in other cases I do because I like the way the decimal point aligns. It makes for easier reading IMO.

```
[[[  1.0001e4], 
  [       1e5]], 

 [[    0.1100], 
  [    0.2690]]]

vs 

[[[  1.0001e4], 
  [       1e5]], 

 [[      0.11], 
  [     0.269]]]

```

Arefaltwir:
ok

AlexDBlack:
OK, so I'm kind of undecided between thowe two cases in your most recent comment... decimal alignment is good for readability, as is removing zeros.
The per-element example you have there is pretty good - I guess my only concern is the fact that we show only 2 significant figures for the 1e-3 case (vs. up to 8 for the 1e3 case).
Otherwise the formatting here is looking pretty good overall, IMO :+1:

raver119:
p.s. i think we might never see those “how to change INDArray precision” messages :)

eraly:
I had looked into printing so that the significant digits are consistent across elements a while back. I didn't find a solution I was happy with. Android Java has some support apparently but not DecimalFormatting. Something to look into later. 

For now I am going to merge. Yay.

eraly:
For the curious on sig digits:
https://stackoverflow.com/questions/5474742/is-there-a-java-number-formatting-library-that-handles-significant-digits

PR #2739, opened 2018-03-14
Increase the maximum heap size for tests with CUDA backend

sshepel:

If maximum heap size less than 6g (rounded value), you may see following kind of errors `Caused by: java.lang.OutOfMemoryError: Physical memory usage is too high: physicalBytes = 3G > maxPhysicalBytes = 3G` from the build log.

## How was this patch tested?
Tested locally and with Jenkins


saudet:
FYI, this also inforces a limit, so we can't have tests that go over that limit.

Issue #2736, opened 2018-03-14
Test coverage for recently added ops


maxpumperla:
In #2730 we've added a bunch ops and a lot of them still need testing.

First of all, some of the transform tests have issues. You'll find some TODOs in the code: https://github.com/deeplearning4j/nd4j/blob/b2667c0eb7cc181195f703b73fed24aa5350ee96/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/autodiff/gradcheck/GradCheckTransforms.java

Most of these worked at some point, but fail now (likely) due to things moving in libnd4j or SameDiff in general.

On top of that, ops that still need testing: 
- [ ] Dilation2D
- [ ] Shape
- [ ] Gather
- [ ] GatherNd
- [ ] Stack
- [ ] Unstack

raver119:
Some of these ops are actually tested in TF import graphs. But definitely, having more  tests never hurts.

PR #2741, opened 2018-03-15
[WIP] Tests for new SameDiff ops

farizrahman4u:
Related:

#2730
#2736 

* Note: Not tested at all.. some issues when building from source, I am stuck with just snapshots for now.


nd4j-backends/nd4j-tests/src/test/java/org/nd4j/autodiff/samediff/SameDiffTests.java
```diff
@@ -276,10 +277,73 @@ public void testTranspose() {
...
+        assertEquals(expected1, result1.eval());
+        assertEquals(expected2, result2.eval());
+
+    }
 
+    @Test
+    public void testGatherNd() {
+        SameDiff sameDiff = SameDiff.create();
+        INDArray arr1 = Transforms.sigmoid(Nd4j.linspace(1, 24, 24)).reshape(2, 3, 4);
+        INDArray arr2 = Nd4j.create(new int[]{1, 2, 3, 0, 1, 3, 1, 0, 2}, new int[]{3, 3});
```
maxpumperla: this won't work. ND4J will think the first int array has shape info. 0 is not allowed. etc.

maxpumperla:
also @farizrahman4u if you could just work from a branch of nd4j from `deeplearning4j` domain, it'd be much easier for me to just check out your work and run the tests (without cloning your fork).

maxpumperla:
@farizrahman4u where are we with this? let me know if you need help with something

farizrahman4u:
@maxpumperla Gather and UnStack are failing. There are older ops failing as well. Let's merge these tests now and fix ops in a different PR.

PR #2742, opened 2018-03-15
<WIP; Donot merge> samediff mapping

eraly:
I am just figuring this out...Will be adding more ops. 
The constructor can get a little crazy with all the options.


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/shape/OneHot.java
```diff
@@ -0,0 +1,94 @@
...
+        attrs.put("off", off);
+
+        ret.put(tensorflowName(),attrs);
+        return ret;
+    }
+
+    @Override
+    public String tensorflowName() {
+        return "OneHot";
+    }
```
maxpumperla: I've been adding 

```java
    @Override
    public String onnxName() {
        throw new NoOpNameFoundException("No onnx name found for shape " + opName());
    }
```

for unavailable ops, but I'm not sure what's better.

maxpumperla:
@eraly one thing I've been (deliberately) omitting is testing actual TF import (to check if property mappings work as proposed), as it's a little expensive to set up. That's something to keep in mind.

eraly:
Agreed. We have larger graphs that contain these ops but I think the better thing to do is write standalone graphs with the op of interest. It is rather rote. Maybe @farizrahman4u and I can split this up.

Steps to check import mapping:
1) Clone https://github.com/deeplearning4j/TFOpTests branch eraly_ops_mapping
2) Change the graph[ here](https://github.com/deeplearning4j/TFOpTests/blob/eraly_ops_mapping/tests/mathops/test_simple_run.py#L38) to include the op we are mapping and run it.
3) Step 2 will write relevant pbs to $DL4J_TEST_RESOURCES/src/main/resources/tf_graphs/examples/simple_run
4) cd to $DL4J_TEST_RESOURCES and run mvn clean install which will get the graph you made into the nd4j testresources
5) Change [modelNames in TFGraphTestList](https://github.com/deeplearning4j/nd4j/blob/025b4fedada833c3dd8bb69cae35ab5d7f42e818/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/imports/TFGraphs/TFGraphTestList.java#L25) to "simple_run" which is the identifier for the graph with the op made in Step 2.
6) Run TFGraphTestList twice for both the libnd4j executioner and the samediff executioner. You just have to switch between the comments[ here](https://github.com/deeplearning4j/nd4j/blob/025b4fedada833c3dd8bb69cae35ab5d7f42e818/nd4j-backends/nd4j-tests/src/test/java/org/nd4j/imports/TFGraphs/TFGraphTestList.java#L76)

So at the end of the day it is writing a simple TF graph in python; running mvn clean install in dl4j-test-resources and running the TFGraphTestList in nd4j and see if it passes.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/shape/OneHot.java
eraly: Added

PR #2766, opened 2018-03-20
Reverse sequence + int 64 TFGraphMapper mapping

eraly:
Fariz's reverse_sequence and the fix needed for int64 byte buffer mapping in TFGraphMapper


eraly:
Import passes with both samediff and libnd4j executor

PR #2777, opened 2018-03-23
testing slice of slice

eraly:
Expanding on edge case found with slice of slice in equality and index out of bounds exception with f order arrays


eraly:
Ping @agibsonccc

PR #2794, opened 2018-03-27
tests need lombok

hvesalai:

Add the required lombok dependency to tests

## How was this patch tested?

By running the tests and observing that they didn't compile without this

### DIFF
--- a/nd4j-backends/nd4j-tests/pom.xml
+++ b/nd4j-backends/nd4j-tests/pom.xml
@@ -101,6 +101,13 @@
 
         <!-- Shaded version of Jackson -->
         <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>${lombok.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>org.nd4j</groupId>
             <artifactId>jackson</artifactId>
             <version>${project.version}</version>


AlexDBlack:
I'm pretty sure this is necessary? I can run the tests via both intellij and maven on master (i.e., adding this dependency isn't required) - and yes, I just re-checked. So I'm closing this unless it can be demonstrated that it's actually required for tests to run.

hvesalai:
@AlexDBlack I think I demonstrated that this is required with mvn `3.5.3`. When you tried, did you have a fresh clone from github to rule out any interference from your earlier builds?

If you want to give me the mvn command to run the tests with, I can run an upload the resulting build logs. All I know is that without this, the tests would not even compile for me. And I think that is natural because the dependency is declared as `provided` in `nd4j-api`.

hvesalai:
The command I was running was both `mvn -Dtest=Nd4jTestsC test -pl :nd4j-tests` on the top level and `mvn -Dtest=Nd4jTestsC test` in the `nd4j-tests` directory. Both started working after this change.

saudet:
@hvesalai You must be doing something wrong, it works just fine for me as well with Maven 3.5.2.

The commands you are using though don't actually do anything. Did you mean `mvn -Dtest=Nd4jTestsC test -Ptestresources` in the `nd4j-native` directory?

hvesalai:
@saudet No, I used the command I provided. The thing I failed to mention was that I edited the nd4j-tests pom.xml to enabled the tests as instructed by @treo 

Anyway, I think I found the reason for this behavior. I was working on top of commit 4cdd4fab4e9434ed9fecd90862d50962e58216c5, so I was missing commit a0e5db60, which also modified the lombok dependencies.

I confirm that this patch is not needed and I can also run the tests without it on top of latest master. I'm sorry for all the trouble.

AlexDBlack:
@hvesalai no worries, not a big deal - thanks for clarifying.

PR #2805, opened 2018-04-01
add SequenceMask op + tests

farizrahman4u:

Add SequenceMask op

## How was this patch tested?

Unit tested in SameDiffTests

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/shape/SequenceMask.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/shape/SequenceMask.java
@@ -1,0 +1,83 @@
+package org.nd4j.linalg.api.ops.impl.shape;
+
+import lombok.NoArgsConstructor;
+import lombok.val;
+import onnx.OnnxMlProto3;
+import org.nd4j.autodiff.samediff.SDVariable;
+import org.nd4j.autodiff.samediff.SameDiff;
+import org.nd4j.imports.NoOpNameFoundException;
+import org.nd4j.imports.descriptors.properties.PropertyMapping;
+import org.nd4j.imports.graphmapper.tf.TFGraphMapper;
+import org.nd4j.linalg.api.ops.DynamicCustomOp;
+import org.tensorflow.framework.AttrValue;
+import org.tensorflow.framework.GraphDef;
+import org.tensorflow.framework.NodeDef;
+
+import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


farizrahman4u:
Note : Merge only after https://github.com/deeplearning4j/libnd4j/pull/849 is merged.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/shape/SequenceMask.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public void initFromTensorFlow(NodeDef nodeDef, SameDiff initWith, Map<String, AttrValue> attributesForNode, GraphDef graph) {
+        TFGraphMapper.getInstance().initFunctionFromProperties(nodeDef.getOp(), this, attributesForNode, nodeDef, graph);
+        addArguments();
+    }
+    @Override
+    public Map<String, Map<String, PropertyMapping>> mappingsForFunction() {
+        Map<String, Map<String, PropertyMapping>> ret = new HashMap<>();
+        Map<String, PropertyMapping> attrs = new LinkedHashMap<>();
+        if (maxLen >= 0) {
```
raver119: How this condition is going to happen during initialization from TF?

All ops during import are initialized with default constructor, so maxLen value will be always -1.

nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/shape/SequenceMask.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public String onnxName() {
+        throw new NoOpNameFoundException("No onnx opName found for " + opName());
+    }
+
+
+    @Override
+    public String[] tensorflowNames() {
```
raver119: You are implementing wrong method here, and with this method implemented only import wont work.

You should be implementing tensorflowName() instead.

raver119: P.s. if not overriden that method throws: https://github.com/deeplearning4j/nd4j/blob/930ad274fd2f8efaf3e73119cc9fb009b8f84571/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/DynamicCustomOp.java#L688-L690

But method tensorflowName() is called at line 64 in your op.

PR #2818, opened 2018-04-03
Fix slice of slice bug

agibsonccc:
Weird corner case found with slicing vectors. This fixes that with tests added by @eraly 
(Please fill in changes proposed in this fix)

## How was this patch tested?

Please review


nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ndarray/BaseNDArray.java
```diff
@@ -6052,6 +6053,12 @@ public INDArray convertToDoubles() {
             return DataBuffer.TypeEx.FLOAT;
         } else if (type == DataBuffer.Type.DOUBLE) {
             return DataBuffer.TypeEx.DOUBLE;
+
+        } else if(type == DataBuffer.Type.INT) {
+            return DataBuffer.TypeEx.INT8;
+        } else if(type == DataBuffer.Type.LONG) {
+            return DataBuffer.TypeEx.INT16;
```
raver119: Nono. INT should be INT32. And LONG should be INT64 if there's INT64.

raver119: i mean number in INT8 stands for number of bits :)

Issue #2831, opened 2018-04-13
NDArrayUtil implementation to convert matrices is wrong


HDouss:
#### Issue Description

`NDArrayUtil.toNDArray(int[][] nums)` implementation is wrong and give bad dimensions. The only place where this methos is called in deeplearning4j, datavec and nd4j is in `ImageLoader.asMatrix` method. To reproduce an issue, run `MnistImagePipelineExample` with these modifications before initializing the record reader:
```java
Configuration conf = new Configuration();
conf.setBoolean(RecordReader.APPEND_LABEL, true);
conf.set("org.datavec.api.records.reader.RecordReader.imageloader", "imageio");
recordReader.initialize(conf ,train);
```
The images read will have bad shapes.

#### Version Information

* version: master branch

saudet:
Thanks for the fix!

PR #2832, opened 2018-04-13
#Fix for 2831

HDouss:

Fixed the implementation.

## How was this patch tested?
It seems to me that nd4j-api does not have unit test. Tested locally by reading images with ImageLoader (scenario described in issue).

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/util/NDArrayUtil.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/util/NDArrayUtil.java
@@ -15,11 +15,11 @@ public class NDArrayUtil {
     public static INDArray toNDArray(int[][] nums) {
         if (Nd4j.dataType() == DataBuffer.Type.DOUBLE) {
             double[] doubles = ArrayUtil.toDoubles(nums);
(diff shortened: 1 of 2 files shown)


agibsonccc:
@HDouss it does, at least ask before making assumptions that a widely used library doesn't have unit tests o_0. Please add your test to: https://github.com/deeplearning4j/nd4j/tree/master/nd4j-backends/nd4j-tests/src/test/java/org/nd4j The reason we have the unit tests in here is because nd4j is a multi backend library. Different implementations should have the same results.

HDouss:
@agibsonccc I did not ask because the method and the modifications are even too simple to test. By the way I did not find any unit test for any method of NDArrayUtil. Could you please point me to the test class, may be I missed something?

agibsonccc:
@HDouss for this one I agree with you..a lot of the simpler classes/utilities that weren't meant for public consumption didn't have tests. if it's impacting someone though we generally ask folks to create tests both for our own maintenance as well as for adding simple cases. Feel free to create a simple test class there under the same package name. It can be used for future reference. Thanks for understanding!

saudet:
Since this affects DataVec, just add the few lines of code you have in #2831 in some unit test there and that'll be alright :) Thanks

HDouss:
@saudet I don't think it is a good idea. I added a unit test covering 2 methods of NDArrayUtil

Issue #2839, opened 2018-04-18
Shade Guava (or provide equivalent Preconditions functionality)


AlexDBlack:
Recently in DL4J I was using some of the Preconditions.checkArguments methods from Guava that were added in version 20: https://github.com/google/guava/wiki/Release20#commonbase

However, (after running into some method not found exceptions) I had to remove said checks (and go back to a more verbose multi-line if statement based checks) as other dependencies were pulling in earlier versions of Guava.

We may not need to shade the entire library; instead, we could provide something equivalent to the Preconditions.checkArgument methods: https://github.com/google/guava/blob/master/android/guava/src/com/google/common/base/Preconditions.java

PR #2845, opened 2018-04-20
Add nd4j-common Preconditions class

AlexDBlack:
Addresses: https://github.com/deeplearning4j/nd4j/issues/2839

This PR adds a Preconditions file with the following methods:
- checkArgument
- checkState
- checkNotNull

We can add other methods if required.

One of the main reasons for this PR: avoiding incompatibility issues.
I've run into a number of NoSuchMethodException issues due to my uses of some Preconditions methods, because it turns out that I'm using a method that was added in say Guava 20, but some other library (Spark, Play, etc) is pulling in say Guava 18. We have enough trouble with transitive dependencies without adding more :)
And there's no way I'm going to be able to consistently remember exactly which Guava Preconditions methods are safe to use and where, given our transitive dependencies and possible lack of control over the classpath with things like Spark submit.

The API is basically identical to Google Guava's Preconditions class (but written from scratch), with some improvements:

**First: "No-boxing" methods for more than 2 primitives**
Why: Preconditions.checkArgument(true, "%s %s %s %s", 1, 2, 3, 4) results in 5 new objects - an Object[] (via varargs method) and 4 Integer arguments (via autoboxing) even though no exception is thrown.
Guava has a few primitive methods like this for the same reason, but are limited to 2 primitives only.
Here: no object creation unless an exception is actually thrown.

**Handles ararys and primitive arrays**
Code: ```Preconditions.checkArgument(false, "Your array: %s", new int[]{1,2,3});```
Guava: ```java.lang.IllegalArgumentException: Your array: [I@1c2c22f3```
Here: ```java.lang.IllegalArgumentException: Your array: [1, 2, 3]```

-----

**Finally, a reminder on proper usage**:
This applies to Guava preconditions as well as ND4J Common Preconditions

Do this:
```Preconditions.checkArgument(x > 0, "x must be >0. Got: x = %s", x)```
Do NOT do this:
```Preconditions.checkArgument(x > 0, "x must be >0. Got: x = " + x)```

Why? The latter creates an object (a new String) that won't be used most of the time, and will then have to be cleaned up by the garbage collector. The former does not create any objects, unless an exception is actually thrown.

### CODE REVIEW

File: nd4j-common/src/main/java/org/nd4j/base/Preconditions.java [hunk 1]
```diff
@@ -0,0 +1,593 @@
...
+        String f = format(message, args);
+        throw new IllegalStateException(f);
+    }
+
+    private static void throwNullPointerEx(String message, Object... args) {
+        String f = format(message, args);
+        throw new NullPointerException(f);
+    }
+
+    private static String format(String message, Object... args) {
```
saudet: Could we use `String.format()` for this instead? Or is it what wasn't working with Guava?

AlexDBlack: This is similar to how Guava does it... even if we handle the arrays etc toString issue before passing args to String.format, the main issue is how it handles (or rather, doesn't) invalid number of args.

Basically String.format ignores the remaining args if you have more than the number of "%s", and throws an exception if you have too few. At least this way, an invalid message gives you something viable instead of a ```java.util.MissingFormatArgumentException: Format specifier '%s'``` exception (which is what String.format gives if you have too few args). tl;dr it's safer in the case of inadventant misuse.

PR #2861, opened 2018-04-27
added InvertMatrix#pinvert based on commons-maths

tteofili:

Add a method to calculate pseudo matrix inverse based on QR matrix decomposition.

## How was this patch tested?

Manual tests in an application to calculate matrix pseudo inverse (for a non squared matrix).


slang03:
How about using `SingularValueDecomposition` instead of `QRDecomposition` as done in [numpy](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.linalg.pinv.html)?

Issue #2858, opened 2018-04-26
add pseudo inverse support


tteofili:
#### Issue Description

It should be possible to do the [pseudo inverse](https://en.wikipedia.org/wiki/Generalized_inverse) of a matrix in ND4J.
That would help in solving systems but also in applications to e.g. transformation matrices for word vectors in machine translation where the no. of rows (dictionary) is much larger than the no. of columns (layer size) and hence plain inverse would not work (as the matrix is not square).

#### Version Information

nd4j 1.0.0-alpha

#### Contributing

I think that could be done by leveraging QR decomposition or SVD. I have a first version I can contribute based on commons-maths (same as InvertMatrix).

saudet:
Please do send a pull request, with an adaption to ND4J, if possible. Thanks!

tteofili:
added a first impl based on commons-maths in PR #2861

saudet:
Although I've merged #2861 and #2883, let's keep this issue opened to track what remains to be done for this feature. Can we say we'll be done after providing an implementation that uses SVD along with any missing tests?

orat:
Is there any progress? I am very interested in using svd in my application.

PR #2883, opened 2018-05-08
Add Java ND4J (Moore-Penrose) Pseudo Inverse (left/right) Implementation

slang03:

Add methods to compute the left and right pseudo-inverse by simple
matrix multiplications directly in ND4J according to
https://en.wikipedia.org/wiki/Moore%E2%80%93Penrose_inverse#Definition

The method makes use of `InvertMatrix.invert(..)` which at the moment
uses apache commons math so this is not a 100% ND4J based implementation but
as soon as someone implements an invert method in ND4J using Blas this
should be fine.

See also: #2858

## How was this patch tested?

Tests for left and right inverse are added in TestInvertMatrices.java
for working and nonworking (singular matrix) cases.

### DIFF
--- a/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/inverse/InvertMatrix.java
+++ b/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/inverse/InvertMatrix.java
@@ -2,9 +2,9 @@ package org.nd4j.linalg.inverse;
 
 import org.apache.commons.math3.linear.LUDecomposition;
 import org.apache.commons.math3.linear.RealMatrix;
+import org.apache.commons.math3.linear.SingularMatrixException;
 import org.nd4j.linalg.api.ndarray.INDArray;
 import org.nd4j.linalg.checkutil.CheckUtil;
-import org.nd4j.linalg.factory.Nd4j;
 
 /**
  * Created by agibsoncccc on 11/30/15.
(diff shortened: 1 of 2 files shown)


saudet:
Is this the same as pull #2861?

slang03:
No, #2861 relies on apache commons math's `QRDecomposition` (See also [wiki](https://en.wikipedia.org/wiki/QR_decomposition#Using_for_solution_to_linear_inverse_problems)) while this pull request offers the direct closed-form solutions to right/left pseudo-inverses.

saudet:
Ah, I see, what do you think @tteofili?

tteofili:
sounds good to me to use this pseudo inverse impl because it only relies on commons-math for the inverse calculation, so as soon as inverse is fixed with a pure nd4j impl, the pseudo inverse calculation won't depend on commons-math anymore either.

slang03:
I'm not sure but it might be worth having both implementations. I'm not too deep into the actual math there, but is there a case where you cannot compute the pseudo matrix via the closed form solution while it is possible using a decomposition method? The closed form solution requires the input matrix to have either full row rank (right pseudo-inverse) or full column rank (left pseudo-inverse). The decomposition methods require the decomposed matrices to be non-singular as they usually use some components of the decomposed matrix and invert them which results in the pseudo-inverse for the input matrix. Is there a situation in which the input matrix does not have full rank but the decomposed matrices are not singular? @tteofili

saudet:
Ok, let's merge both for now and keep on working on this :) We'll need to add a test for `pinvert()` too...

Thanks for the contributions guys!

PR #2899, opened 2018-05-14
#5143 add enforcer profile

AlexDBlack:
Addresses: https://github.com/deeplearning4j/deeplearning4j/issues/5143

Should be good to go, want to run some final tests before merging...


AlexDBlack:
@sshepel Thanks
I was thinking we hold off merging until after release (but before mono-repo) - just in case we've missed something here.

